PR #138, opened 2015-01-09
Fix: Atmel Studio Project for samr21_xpro/03oos_openwsn

saurabh984:
Added RRT Dependency
Fixed Compiler warnings for RRT
Updated CMSIS (Stack Size Correction)
ASF 3.20.1


xvilajosana:
saurabh984, 

I've seen that you committed a binary file named 
projects/samr21_xpro/03oos_openwsn/03oos_openwsn.atsuo

is this a required file for the avr studio? is this a configuration file or something that avr studio generates from other configuration files? 

If this file is autogenerated I would remove it. If it is required, no worries, should stay there.

thanks!
X

PR #143, opened 2015-02-24
Develop fw 335

NicolaAccettura:
Fix.


drivers/common/openserial.c
```diff
@@ -200,6 +205,10 @@ owerror_t openserial_printCritical(uint8_t calling_component, uint8_t error_code
    );
 }
 
+void openserial_board_reset_cb(opentimer_id_t id) {
+   board_reset();
+}
```
twatteyne: good catch!

twatteyne:
Excellent, thanks!

PR #156, opened 2015-03-18
FW-348. Crypto API test suite and example project.

malishav:
This pull request introduces a board independent project to demonstrate the use of ```CRYPTO_ENGINE``` and runs a test suite with some predefined test vectors. In case all tests passed successfully the node turns on the radio LED and hangs. If there was a single error, the node turns on the error LED.


twatteyne:
:100:
:+1: 
:tada:

PR #158, opened 2015-03-18
FW-353. Fix raw CBC encryption.

malishav:
This pull request fixes the bug in firmware implementation of CBC encryption.


twatteyne:
Awesome!! :hamburger:

PR #165, opened 2015-04-02
FW-367. Fix.

NicolaAccettura:


openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -171,6 +171,9 @@ void isr_ieee154e_newSlot() {
    if (ieee154e_vars.isSync==FALSE) {
       if (idmanager_getIsDAGroot()==TRUE) {
          changeIsSync(TRUE);
+         incrementAsnOffset();
```
twatteyne: why increment?

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -715,6 +718,13 @@ port_INLINE bool ieee154e_processIEs(OpenQueueEntry_t* pkt, uint16_t* lenIE) {
             // at this point, ASN and frame length are known
             // the current slotoffset can be inferred
             ieee154e_syncSlotOffset();
+            schedule_syncSlotOffset(ieee154e_vars.slotOffset);
```
twatteyne: why pass this param?

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -715,6 +718,13 @@ port_INLINE bool ieee154e_processIEs(OpenQueueEntry_t* pkt, uint16_t* lenIE) {
             // at this point, ASN and frame length are known
             // the current slotoffset can be inferred
             ieee154e_syncSlotOffset();
+            schedule_syncSlotOffset(ieee154e_vars.slotOffset);
+            ieee154e_vars.nextActiveSlotOffset = schedule_getNextActiveSlotOffset();
+            /* 
+            infer the asnOffset based on the fact that
+            ieee154e_vars.freq = 11 + (asnOffset + channelOffset)%16 
+            */
+            ieee154e_vars.asnOffset = ieee154e_vars.freq - 11 - schedule_getChannelOffset();
```
twatteyne: wow, this is a dependent we don't want.
What is I want to run the network on a single frequency, or change my hopping sequence?

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -1705,8 +1717,14 @@ port_INLINE void incrementAsnOffset() {
...
       }
    }
+   
    // increment the offsets
-   ieee154e_vars.slotOffset  = (ieee154e_vars.slotOffset+1)%schedule_getFrameLength();
+   frameLength = schedule_getFrameLength();
+   if (frameLength != 0) {
+      ieee154e_vars.slotOffset  = (ieee154e_vars.slotOffset+1)%frameLength;
+   } else {
+      ieee154e_vars.slotOffset  = ieee154e_vars.slotOffset + 1;
```
twatteyne: why not simply initialize to a certain value? why do ++?

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -1705,8 +1717,14 @@ port_INLINE void incrementAsnOffset() {
          ieee154e_vars.asn.byte4++;
       }
    }
+   
    // increment the offsets
-   ieee154e_vars.slotOffset  = (ieee154e_vars.slotOffset+1)%schedule_getFrameLength();
+   frameLength = schedule_getFrameLength();
+   if (frameLength != 0) {
```
twatteyne: please invert the two statements and use `f (frameLength==0) {`

openstack/02a-MAClow/IEEE802154E.c
NicolaAccettura: Just for uniformity and to account for all timeslots elapsed: earlier, when the mote had detected to have been configured as a DAGroot (by openvisualizer) in a given timeslot, it would have changed to sync without incrementing the ASN in that timeslot. Now it does. So we can keep track of all the timeslots really elapsed from the beginning (the beginning is defined as ASN=0 on the mote that later has been configured as DAGroot).

openstack/02a-MAClow/IEEE802154E.c
NicolaAccettura: ieee154e_syncSlotOffset() is in charge of matching the slotoffset to the ASN according to the slotframe length, so it updates ieee154e_vars.slotOffset. That parameter is then passed to schedul_syncSlotOffset, which instead matches the slotOffset to the schedule.

openstack/02a-MAClow/IEEE802154E.c
NicolaAccettura: This function has just been moved around from elsewhere, it was already something running. The channel hopping sequence is defined in calculateFrequency. This function is triggered only when a mote receives EBs. Should be ieee154e_vars.asnOffset be (ASN % 16) as in incrementAsnOffset() ?

openstack/02a-MAClow/IEEE802154E.c
NicolaAccettura: ieee154e_vars.slotoffset is used in many prints to openserial: at the beginning (when a mote is not synchronized) the slotoffset is simply incremented as the ASN; of couse, the value of the slotOffset will wrap around to 0 after about 16 minutes with a 15ms timeslot (for this reason, when a mote is set to be DAGroot by openvisualizer, the slotOffset is recomputed by ieee154e_syncSlotOffset on the basis of the ASN which wraps around slowlier), but we can see errors and maybe relate them in time.
++ has been done.

openstack/02a-MAClow/IEEE802154E.c
NicolaAccettura: Done.

twatteyne:
@NicolaAccettura, thanks for fixing this, and providing answers. I'll merge now.

PR #166, opened 2015-04-04
FW-369. Create a CC2420-specific library from existing code.

malishav:
This pull request introduces cc2420.c file within ```bsp/chips/cc2420``` and moves the private functions from ```radio.c``` file into this file. 

The only modification to the existing code is in the first commit where ```radio_spiReadRxFifo()``` used to directly reference the global variable ```radio_vars_t radio_vars``` defined in ```radio.c``` file. This function is only called from ```radio_getReceivedFrame()``` with the first argument set to the pointer of ```radio_vars```. Thus, this pull request replaces the direct reference to the ```radio_vars``` in ```radio_spiReadRxFifo()``` with the function parameter ```statusRead``` in order to have clear separation between the ```radio.c``` and ```cc2420.c``` files.


twatteyne:
Wonderful! :+1:

PR #167, opened 2015-04-09
FW-240. Security Extension - TelematicsLab

ssciancalepore:
Integration of security processing and operations within the OpenWSN protocol stack


inc/opendefs.h
```diff
@@ -158,6 +158,9 @@ enum {
    COMPONENT_TOHLONE                   = 0x22,
    COMPONENT_UECHO                     = 0x23,
    COMPONENT_RRT                       = 0x24,
+   //START OF TELEMATICS CODE
```
twatteyne: don't put comments where you start/stop.
If everyone did this, it would be a nightmare. It's important your name appears as a contributor: in each file you changes, just add you name in the description of the file on the top

inc/opendefs.h
```diff
@@ -293,6 +303,24 @@ typedef struct {
    uint8_t       l2_joinPriority;                // the join priority received in EB
    bool          l2_IEListPresent;               //did have IE field?
    bool          l2_joinPriorityPresent;
+   //START OF TELEMATICS CODE
+   //security
+   bool			 	 l2_security;				 //flag for security enabled/disabled
```
twatteyne: please respect indendation and alignment.
Also, please configure your editor to not include any tabs. We don't want a single tab anywhere. tabs are evil.

openstack/02a-MAClow/IEEE802154.c
```diff
@@ -34,6 +34,14 @@ void ieee802154_prependHeader(OpenQueueEntry_t* msg,
    uint8_t temp_8b;
    
    //General IEs here (those that are carried in all packets) -- None by now.
+   //START OF TELEMATICS CODE
+   /*if security is enabled, the Auxiliary Security Header need to be
```
twatteyne: comments with `//`, always. Allows us to comment out entire sections

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -171,6 +175,10 @@ void isr_ieee154e_newSlot() {
    if (ieee154e_vars.isSync==FALSE) {
       if (idmanager_getIsDAGroot()==TRUE) {
          changeIsSync(TRUE);
+	 //START OF TELEMATICS CODE
+         //If DAG Root, store the Key
+	 scheduler_push_task(coordinatorORParent_init,TASKPRIO_SIXTOP);
```
twatteyne: if you push a task, you don't know when it will execute. I haven't read hwat `coordinatorORParent_init()` does, but make sure it is not accessing any variable that could have been changed by the time it executes

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -535,6 +543,14 @@ port_INLINE void activity_synchronize_endOfFrame(PORT_RADIOTIMER_WIDTH capturedT
       
       // parse the IEEE802.15.4 header (synchronize, end of frame)
       ieee802154_retrieveHeader(ieee154e_vars.dataReceived,&ieee802514_header);
+
+      //START OF TELEMATICS CODE
+      //if not DAG Root and not synch, store the Key
+  	  if(idmanager_getIsDAGroot()==FALSE && ieee154e_isSynch() == FALSE
+  		 && ieee802514_header.frameType == IEEE154_TYPE_BEACON){
+  		  remote_init(ieee802514_header);
```
twatteyne: `remote_init()` is a totally obscure name. Are you initializing your remote controlled airplane?
Please prefix all function name by the file it's in.

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -877,6 +893,27 @@ port_INLINE void activity_ti1ORri1() {
...
+             * if security is enabled on the current frame,
+             * I have to encrypt it before sending.
+             * Note that if this is not the first attempt, the
+             * clearText has to be encrypted again.
+             */
+            if(ieee154e_vars.dataToSend->l2_security == IEEE154_SEC_YES_SECURITY){
+            	//if it is a retransmission, recover the clearText packet
+            	if(ieee154e_vars.dataToSend->l2_numTxAttempts != 1){
+					 ieee154e_vars.dataToSend->length = ieee154e_vars.dataToSend->clearText_length;
+					 uint8_t i;
```
twatteyne: only declare variables at the top of a function

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -877,6 +893,27 @@ port_INLINE void activity_ti1ORri1() {
...
+					 for(i=0;i<ieee154e_vars.dataToSend->length; i++){
+						ieee154e_vars.dataToSend->l2_payload[i] = ieee154e_vars.dataToSend->clearText[i];
+					 }
```
twatteyne: tabs are evil, remove everywhere

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -877,6 +893,27 @@ port_INLINE void activity_ti1ORri1() {
...
+				 }
+
+            	security_outgoingFrame(ieee154e_vars.dataToSend);
```
twatteyne: "secutiry" is a very general word. Would there be something more specific. `ccm_poipoi`?

openstack/02a-MAClow/IEEE802154E.h
```diff
@@ -127,16 +127,16 @@ typedef enum {
 //    - duration_in_seconds = ticks / 32768
 enum ieee154e_atomicdurations_enum {
    // time-slot related
-   TsTxOffset                =  131,                  //  4000us
+   TsTxOffset                =  131+160,                  //  4000us
```
twatteyne: wow, why those changes?
In any case, don't do a `+`, just put the final number

openstack/02a-MAClow/TI_aes.c
```diff
@@ -0,0 +1,221 @@
+
```
twatteyne: add the license for this code somewhere, as this is taken from TI

openstack/02a-MAClow/ccmstar.c
```diff
@@ -0,0 +1,434 @@
+/**
+\brief CCMstar software procedure
+
+\author Savio Sciancalepore <savio.sciancalepore@poliba.it>, March 2015.
+*/
+
+#include "ccmstar.h"
+
+void CCMstar(OpenQueueEntry_t* 		pkt,
```
twatteyne: all function names should start with `ccmstar_`.

openstack/02a-MAClow/ccmstar.c
```diff
@@ -0,0 +1,434 @@
...
+}
+
+
+void Input_Transformation(uint8_t* payload,
+		                  uint8_t  length,
+		                  uint8_t authDataLength,
+		                  uint8_t* authData,
+		                  uint8_t* aData){
+
+	// La || a || 0 || m || 0
```
twatteyne: what is this? Please add more comments, or delete

openstack/02a-MAClow/ccmstar.c
```diff
@@ -0,0 +1,434 @@
...
+	uint8_t in[16],out[16];
+	uint8_t i;
+
+	//determine Flags Field
+	B[0] = 0;
+	B[0] = B[0] <<1;//1b reserved
+	if(length == 0){
+		B[0]= B[0] <<1;
+	}
+	else{
```
twatteyne: `} else {`

openstack/02a-MAClow/security.c
```diff
@@ -0,0 +1,798 @@
+/**
+\brief General Security Operations
```
twatteyne: name of this file is too generic

openstack/02a-MAClow/security.c
```diff
@@ -0,0 +1,798 @@
...
+
+	CCMstar(msg,key,nonce);
+
+}
+
+void retrieve_AuxiliarySecurityHeader(OpenQueueEntry_t*      msg,
+                  	  	  	  	      ieee802154_header_iht* tempheader){
+
+	//a check if security is enabled, for me it not useful. If I'm here, security is enabled.
+//	if(tempheader->securityEnabled == TRUE){
```
twatteyne: remove unused code

openstack/02a-MAClow/security.c
```diff
@@ -0,0 +1,798 @@
...
+		}
+
+	//Frame Counter (ASN)
+	uint8_t myASN[5];
+	ieee154e_getAsn(myASN);
+	for(i=0;i<5;i++){
+		nonce[8+i] = myASN[i];
+	}
+
+//	asn_t now;
```
twatteyne: remove unused code

openstack/02a-MAClow/security.h
```diff
@@ -0,0 +1,163 @@
...
+
+typedef struct{
+	m_securityLevelDescriptor     SecurityDescriptorEntry[5];
+	} m_macSecurityLevelTable;
+
+
+//=========================== variables =======================================
+
+
+BEGIN_PACK
```
twatteyne: why pack?

openstack/02b-MAChigh/sixtop.c
```diff
@@ -462,14 +477,25 @@ void task_sixtopNotifReceive() {
...
-         }
-         break;
+                 if (msg->length>0) {
+         	 //START OF TELEMATICS CODE
+         	 //discard packets with security problems
+         	 if(msg->l2_toDiscard == 0){
+         		 // send to upper layer
+         		 iphc_receive(msg);
+         	 } else {
+ 			   openserial_printError(COMPONENT_SECURITY,ERR_SECURITY,
```
twatteyne: this is `COMPONENT_SIXTOP` speaking

twatteyne:
Excellent! Please see my comments inline. Also, please make sure the Travis build passes.

twatteyne:
@ssciancalepore, @malishav, how do you suggest we move on with this PR?

Now that Malisa is done (almost, right?) with the low-level stuff, it might be time to merge this one in. It will have been the longest open PR in OpenWNS history! Sorry, @ssciancalepore.

@malishav, do you want to take a stab at doing the merging exercise?

Thomas

malishav:
@twatteyne, yes, I will run some tests next week. @ssciancalepore, in case I have some changes to commit, I was thinking of forking your repo and then submitting a PR which you could review. What do you think?

twatteyne:
Wonderful, thanks!

ssciancalepore:
@twatteyne @malishav Ok for me. 

In the last commits I integrated suggestions by Thomas and the crypto engine from Malisa, and tested the solution on OpenMote-CC2538. It works. Problems on Travis CI build are related to RAM footprint for Z1 and gina boards and building problems on python board.

As regarding python board, I integrated the module variables, functions and header files in Sconscript.env in the python project, but I still have problems with the module global variable.

@malishav I completely agree.

twatteyne:
Wonderful, congrats on the super progress!

If there the implementation exceeds memory on some boards, we should make sure the default behavior of the build system for those board is to have no security.

Global variables of each module should be grouped into a struct, and that struct declared in the SCons build systen for Python.

@malishav, is this something you could take a look at?

malishav:
Yeap!

malishav:
@ssciancalepore, thanks for your time today and for answering all the questions! Some questions/remarks after a (new) pass over the code:

-  You should check the return value of ```CRYPTO_ENGINE``` in order to make sure that frames that fail security processing are not passed to the higher layer. Code should handle the case where something went wrong and ```CRYPTO_ENGINE``` returns ```E_FAIL```. Also, ```outgoing``` and ```incoming``` security routines should return ```owerror_t``` in order to signal if something went wrong to the higher layer.
- I am not clear how you define the length of Associated Data to be passed to ```CRYPTO_ENGINE```. Right now, the argument that is passed is ```msg->length - msg->l2_length```. Does this take into account the presence of header IE that is only authenticated and/or payload IEs that should be encrypted? I think there should be logic somewhere that should differentiate between different types of frames (```DATA```, ```ACK```, ```BEACON```, ```COMMAND```) and what is encrypted/authenticated for each, depending on IEs present. Ideally, this should be handled inside the ```outgoing``` and ```incoming``` routines.
- Variable ```l2_security``` in ```opendefs.h``` can be removed and we could instead use ```l2_securityLevel != ASH_SLF_TYPE_NOSEC``` condition.
- I am working on removing ```aData[]``` and ```clearText[]``` arrays from ```OpenQueueEntry_t``` in order to encrypt/authenticate frames without having to keep these arrays for each packet. For now, my modifs are all in ```IEEE802154E.c``` but please let me know once you push your latest commits as I would also need to adapt the code around ```CRYPTO_ENGINE``` calls. For now, I am focusing on the entry points of the security code within the ```IEEE802154E.c``` module. 

Thanks for all the work!!!

twatteyne:
Fantastic!
Please note that the Travis build fails for the Python simulator. Probably a quick fix?
When you get a chance, could you also resync your fork as an automatic merge isn't possible.
Thanks!

ssciancalepore:
@malishav @twatteyne  

I've just pushed the new version of the code, with some fixes that you suggest me.

1) Now the procedures security_outgoing and security_incoming return an owerror_t variables, which can be used to identify errors related with security processing;

2) associated Data and input Data for the crypto engine have been slightly modified in order to allow for the complete authentication of the packet. The problem with IEs is the next stuff that I have to solve.

3) variable l2_security has been removed, as well as other variables related to the packet (i.e., l2_toDiscard, l2_auxiliaryLength, and aData).

4) aData now have been removed from the packet. I have not touched the clearText[] variable, because I think that some fixes on IEEE802.15.4E module need to be done.

The next stuffs I have to include, probably in next pushes, will be the adding of authentication services on beacon frames and the managing of IEs in the frame.

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,887 @@
...
+      //here I have to insert the ASN: I can only reserve the space and
+      //save the pointer. The ASN will be added by IEEE802.15.4e
+
+      // reserve space
+      packetfunctions_reserveHeaderSize(msg,sizeof(macFrameCounter_t));
+
+      // Keep a pointer to where the ASN will be
+      // Note: the actual value of the current ASN will be written by the
+      //    IEEE802.15.4e when transmitting
+      msg->l2_ASNpayload = msg->payload;
```
changtengfei: @ssciancalepore ,

In syncIE, l2->ASNpayload is also used: https://github.com/ssciancalepore/openwsn-fw/blob/develop/openstack/02b-MAChigh/processIE.c#L60
An other pointer to the ASNpayload  for framecounter should be used.

ssciancalepore: @changtengfei thank you for your finding! 

I will update the code as soon as possible.

malishav:
whohooo green asterix! now we only need to get it working once again ! :)

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,905 @@
...
+                                         msg->l2_ASNpayload);
+   } //otherwise the frame counter is not in the frame
+
+   //nonce creation
+   memset(&nonce[0], 0, 13);
+   switch (msg->l2_keyIdMode){
+      case 0:
+      case 1:
+         for (i=0; i<8; i++){
+            nonce[i] = ieee802154_security_vars.m_macDefaultKeySource.addr_64b[i];
```
malishav: @ssciancalepore, 15.4 nonce should be created from extended/short address independent of keyIdMode

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,905 @@
...
+     return E_FAIL;
+   }
+
+   //create nonce
+   memset(&nonce[0], 0, 13);
+   switch (msg->l2_keyIdMode){
+      case 0:
+      case 1:
+         for (i=0; i<8; i++){
+            nonce[i] = ieee802154_security_vars.m_macDefaultKeySource.addr_64b[i];
```
malishav: same remark for nonce creation here, you should use extended/short address from source address field in the frame..

malishav:
@ssciancalepore, @changtengfei, @twatteyne I tested today the synced security branch with the latest software version. @changtengfei, it seems that there are duplicates of certain packets in ```openwsn-berkeley/develop``` code -- it is easy to notice double DAOs processed in the visualizer. 

Status:
- With security disabled I did not notice any regressions. 

- With security enabled, there is a corruption occurring at L3 with these duplicate packets which causes all sorts of ugly errors in the visualizer. The 'duplicates' *pass* the security check at L2 but once they are processed by IPHC it seems they are corrupted. I did not manage to trace today where these packets are coming from, @changtengfei do you happen to know?

changtengfei:
@malishav , do you mean you see double DAOs were processed in in the code of visualizer, or you see it from the output from visualizer?

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,860 @@
...
+   if (len_a + len_m > 125) {
+      openserial_printError(COMPONENT_SECURITY,ERR_SECURITY,
+                           (errorparameter_t)msg->l2_frameType,
+                           (errorparameter_t)1);
+      return E_FAIL;
+   }
+
+   if (msg->l2_authenticationLength != 0) {
+      //update the length of the packet
+      packetfunctions_reserveFooterSize(msg,msg->l2_authenticationLength);
```
changtengfei: @ssciancalepore @malishav , in the definition of OpenQueueEntry_t,  The last three type bytes are reserved for 2 byte crc and 1 lqi originally. As here we need append the authentication data at end, we may require to extend the size of packet buffer at https://github.com/ssciancalepore/openwsn-fw/blob/develop/inc/opendefs.h#L322. Make sense?

malishav: Not really, authentication tag must fit within the 125 bytes of PHY payload, we cannot extend maxPHYPacketSize from 15.4 std.

malishav:
I see double DAOs processed in the output of the visualizer AND the output of the DAGroot when debugging using serial port.

openstack/02a-MAClow/IEEE802154_security.c
changtengfei: I see. So we need reserve authentication data inside of the 125 bytes of PHY payload. For this code, https://github.com/ssciancalepore/openwsn-fw/blob/develop/openstack/cross-layers/openqueue.c#L246
We need set the payload pointer to 127-4 if the authentication data length is 4. Make sense?

malishav: I don't think so, it is the same case as CRC. You append the authentication tag and CRC to the end of your frame. At this time you don't know what is the length of the frame you will be using...

ssciancalepore: @changtengfei I think it's exactly as @malishav wrote. The procedure it's the same as reserving space for CRC at the end of the frame. The crypto engine will insert the MIC in the last bytes of the MAC payload, indipendently of which authentication tag length you will use

changtengfei: What I concerned is the range of payload buffer. As the size of payload buffer is 1+1+125+2+1(130), the authentication  data starts at 127, when writing the authentication data, the last byte will out of 130bytes reserved, also for later CRC. It may have a chance that the bytes out of payload buffer will be used for other code.

changtengfei:
@malishav @ssciancalepore I did a tiny modification on openwsn-berkeley/openwsn-fw branch. Please pull and merge it in your branch. This will solve the duplicated packet issue. Thanks!

openstack/02a-MAClow/IEEE802154_security.c
malishav: I see your point. Could you check if commit "Shift payload entry point for length of the authentication tag." (d7fdce8bf903) addresses what you had in mind?

Note that latest commits will appear in this PR as soon as @ssciancalepore merges the PR. Until then you can see it here: ```https://github.com/malishav/openwsn-fw/commit/d7fdce8bf90385a66b1bf5189f22e5ed4cda5061```

malishav:
@changtengfei, confirming that the issue of duplicates has been solved with ```https://github.com/openwsn-berkeley/openwsn-fw/commit/f4050d0bfe77e9c0a40386b95364f3b69d7c1de1```. Thanks!

malishav:
@ssciancalepore, @twatteyne, @changtengfei: I just pushed fixes that address all remaining issues, including memory corruption, that I am aware of. These will appear here once @ssciancalepore merges them in. I also tested with ping6 and there do not seem to be any regressions with security *enabled*, in respect to the case without security.

@twatteyne, could you take a look and see if something else should be addressed before this PR is merged? On my side, it is ready to fly. I am ready to fly as well :-).

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -550,11 +551,24 @@ port_INLINE void activity_synchronize_endOfFrame(PORT_RADIOTIMER_WIDTH capturedT
       ieee154e_vars.dataReceived->l2_frameType = ieee802514_header.frameType;
       ieee154e_vars.dataReceived->l2_dsn       = ieee802514_header.dsn;
       memcpy(&(ieee154e_vars.dataReceived->l2_nextORpreviousHop),&(ieee802514_header.src),sizeof(open_addr_t));
-      
-      // toss the IEEE802.15.4 header -- this does not include IEs as they are processed 
+
+      // This is the case where we first hear an EB and do not yet know the ASN used
```
twatteyne: Wow. An EB contains the ASN. We should NOT accept is as-is,  but rather read the ASN from the EB, then authenticate with that information.

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -942,8 +956,26 @@ port_INLINE void activity_ti1ORri1() {
 }
 
 port_INLINE void activity_ti2() {
+   static OpenQueueEntry_t local_copy_for_transmission; // keep in BSS to prevent stack overflow
```
twatteyne: I understand the use of static`

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -1536,21 +1582,36 @@ port_INLINE void activity_ri6() {
...
+   // that were present in the DATA packet.
+   ieee154e_vars.ackToSend->l2_securityLevel = ieee154e_vars.dataReceived->l2_securityLevel;
+   ieee154e_vars.ackToSend->l2_keyIdMode     = ieee154e_vars.dataReceived->l2_keyIdMode;
+   ieee154e_vars.ackToSend->l2_keyIndex      = ieee154e_vars.dataReceived->l2_keyIndex;
+
    ieee802154_prependHeader(ieee154e_vars.ackToSend,
                             ieee154e_vars.ackToSend->l2_frameType,
                             TRUE,//ie in ack
-                            IEEE154_SEC_NO_SECURITY,
+                            ieee154e_vars.ackToSend->l2_securityLevel == IEEE154_ASH_SLF_TYPE_NOSEC ? 0 : 1,
```
twatteyne: why not pass `ieee154e_vars.ackToSend->l2_securityLevel` directly and have the `ìf` statement in the body of `ieee802154_prependHeader()`?

openstack/02a-MAClow/IEEE802154_dummy_security.c
```diff
@@ -0,0 +1,43 @@
+/**
+\brief Dummy IEEE802154 security implementation
```
twatteyne: please specify when this is used in this header.

openstack/02a-MAClow/IEEE802154_dummy_security.h
```diff
@@ -0,0 +1,15 @@
+/**
+\brief Dummy IEEE802154 security implementation header
```
twatteyne: please specify when this is used in this header.

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
+/**
+\brief Security operations defined by IEEE802.15.4e standard
+
+\author Savio Sciancalepore <savio.sciancalepore@poliba.it>, April 2015.
+\author Giuseppe Piro <giuseppe.piro@poliba.it>,
```
twatteyne: add dates

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+/**
+\brief Initialization of security tables and parameters.
+*/
+void IEEE802154_security_init(void){
+
+   //Setting UP Phase
+
+   //MASTER KEY: OpenWSN
+   memset(&ieee802154_security_vars.M_k[0], 0, 16);
+   ieee802154_security_vars.M_k[0] = 0x4e;
```
twatteyne: why `NSWnepo`? According to https://tools.ietf.org/html/draft-ietf-6tisch-minimal-08, K1 should be `36 54 69 53 43 48 6D 69 6E 69 6D 61 6C 31 35` (`6TiSCHminimal15`). For K2, the "tradition" is to use `deadbeeffacecafedeadbeeffacecafe`

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+                                                          frameCounterSuppression,
+                                                          ieee802154_security_vars.m_macFrameCounterMode); //length of Key ID field
+
+
+   if ((msg->length+auxiliaryLength+msg->l2_authenticationLength+2) >= 128 ){ //2 bytes of CRC, 127 MaxPHYPacketSize
+      return;
+   }
+
+   //start setting the Auxiliary Security Header
+   if (msg->l2_keyIdMode !=0){//if the KeyIdMode is zero, keyIndex and KeySource are omitted
```
twatteyne: `0` is a magic value here, please replace by `#define`

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+      temp8b = msg->l2_keyIndex; //key index field
+      packetfunctions_reserveHeaderSize(msg, sizeof(uint8_t));
+      *((uint8_t*)(msg->payload)) = temp8b;
+   }
+
+   switch (msg->l2_keyIdMode){
+      case 0: //no KeyIDMode field - implicit
```
twatteyne: no magic values. replace by `#define`

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+    	 temp_keySource = &ieee802154_security_vars.m_macDefaultKeySource;
+         memcpy(&(msg->l2_keySource), temp_keySource, sizeof(open_addr_t));
+         break;
+      case 1:// macDefaultKeySource
```
twatteyne: no magic values. replace by `#define`

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+         msg->l2_keySource = ieee802154_security_vars.m_macDefaultKeySource;
+         break;
+      case 2: //keySource with 16b address
```
twatteyne: no magic values. replace by `#define`

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+         temp_keySource = &msg->l2_keySource;
+         packetfunctions_reserveHeaderSize(msg, sizeof(uint8_t));
+         *((uint8_t*)(msg->payload)) = temp_keySource->addr_64b[6];
+         packetfunctions_reserveHeaderSize(msg, sizeof(uint8_t));
+         *((uint8_t*)(msg->payload)) = temp_keySource->addr_64b[7];
+         break;
+      case 3: //keySource with 64b address
```
twatteyne: no magic values. replace by `#define`

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+         temp_keySource = &msg->l2_keySource;
+         packetfunctions_writeAddress(msg,temp_keySource,OW_LITTLE_ENDIAN);
+         break;
+      default:
+         return;
```
twatteyne: this should be an error, right?

openstack/02a-MAClow/IEEE802154E.c
malishav: Yes, this is good to separate network with different keys. Note that the comment relates to the implicit ASN. EB contains the explicit ASN which we can use but it still does not provide us with any guarantee against replay attack. The problem here is that you should parse the EB to get ASN before you have it passed the security check, which normally one shouldn't do. I will see to add a join-specific call to get this done, as discussed on the 6tisch list...

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+
+   msg->l2_authenticationLength = IEEE802154_security_authLengthChecking(msg->l2_securityLevel);
+
+   //retrieve the KeyIdMode field
+   msg->l2_keyIdMode = (temp8b >> IEEE154_ASH_SCF_KEY_IDENTIFIER_MODE)& 0x03;//2b
+
+   frameCnt_Suppression = (temp8b >> IEEE154_ASH_SCF_FRAME_CNT_MODE)& 0x01;//1b
+   frameCnt_Size = (temp8b >> IEEE154_ASH_SCF_FRAME_CNT_SIZE)& 0x01;//1b
+
+   tempheader->headerLength = tempheader->headerLength+1;
```
twatteyne: `++`?

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+   if (frameCnt_Suppression == 0){//the frame counter is here
+      //the frame counter size can be 4 or 5 bytes
+      for (i=0;i<frameCnt_Size;i++){
+          receivedASN[i] = *((uint8_t*)(msg->payload)+tempheader->headerLength);
+          tempheader->headerLength = tempheader->headerLength+1;
+      }
+
+      l2_frameCounter.bytes0and1 = receivedASN[0]+256*receivedASN[1];
+      l2_frameCounter.bytes2and3 = receivedASN[2]+256*receivedASN[3];
+      if(frameCnt_Size == 0x04){
```
twatteyne: comment what `0x04` means

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+         l2_frameCounter.byte4 = receivedASN[4];
+      }
+
+      if (l2_frameCounter.byte4 == 0xff){
```
twatteyne: comment what `0xff` means

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+         return;
+      }
+
+   }
+
+   //retrieve the Key Identifier field
+   switch (msg->l2_keyIdMode){
+      case 0:
```
twatteyne: no magic values, applies to all cases

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+         //key is derived implicitly
+         temp_addr = &ieee802154_security_vars.m_macDefaultKeySource;
+         memcpy(&(msg->l2_keySource), temp_addr, sizeof(open_addr_t));
+         break;
+      case 2:
+      packetfunctions_readAddress(((uint8_t*)(msg->payload)+tempheader->headerLength),
```
twatteyne: indentation

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+                                     &msg->l2_keySource,
+                                     OW_LITTLE_ENDIAN);
+         tempheader->headerLength = tempheader->headerLength+8;
+         break;
+      default:
+         return;
+      }
+
+   //retrieve the KeyIndex
+   if (msg->l2_keyIdMode != 0){
```
twatteyne: magic value

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+
+   if (keyDescriptor==NULL){
+      //can't find the key
+      openserial_printError(COMPONENT_SECURITY,ERR_SECURITY,
+                           (errorparameter_t)msg->l2_frameType,
+                           (errorparameter_t)2);
+      return E_FAIL;
+   }
+
+   //g device descriptor lookup
```
twatteyne: why `g`?

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+   if (securityLevelDescriptor == NULL){
+	      //can't find the frame type
+	      openserial_printError(COMPONENT_SECURITY,ERR_SECURITY,
+	                           (errorparameter_t)msg->l2_frameType,
+	                           (errorparameter_t)4);
+	      return E_FAIL;
+   }
+
+   //i+j+k
+   if (IEEE802154_security_incomingSecurityLevelChecking(securityLevelDescriptor,
```
twatteyne: I'm not a big fan of this construction, as the core of the behavior is hidden in the `if` statement.

I recommend you create an `outcome` variable, us that to hold the rc of `IEEE802154_security_incomingSecurityLevelChecking()`, then to the `if` on the `outcome` variable.

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+//=========================== private =========================================
+
+/**
+\brief Identification of the length of the MIC.
+*/
+uint8_t IEEE802154_security_authLengthChecking(uint8_t securityLevel){
+
+   uint8_t authlen;
+   switch (securityLevel) {
+      case 0 :
```
twatteyne: magic values

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+
+//=========================== private =========================================
+
+/**
+\brief Identification of the length of the MIC.
+*/
+uint8_t IEEE802154_security_authLengthChecking(uint8_t securityLevel){
+
+   uint8_t authlen;
+   switch (securityLevel) {
```
twatteyne: You might group all `securityLevel` options with the same `authLen`

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+         frameCntLength = 4;
+      } else {
+         frameCntLength = 5;
+      }
+   } else {
+      frameCntLength = 0;
+   }
+
+   switch (KeyIdMode){
+      case 0:
```
twatteyne: magic values...

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+         auxilary_len = frameCntLength + 1 + 9; //SecLev, FrameCnt, KeyIdMode (8 bytes) and KeyIndex
+         break;
+      default:
+         break;
+    }
+   return auxilary_len;
+}
+
+/**
+\brief Verify if the key used to protect the frame has been used properly.
```
twatteyne: this comment contains 0 information. What does "used properly" mean?

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+         return TRUE;
+      }
+   }
+
+   ENABLE_INTERRUPTS();
+   return FALSE;
+}
+
+/**
+\brief Verify if the Security Level of the incoming Frame is acceptable or not.
```
twatteyne: same here, what does "acceptable" mean?

openstack/02a-MAClow/IEEE802154_security.c
```diff
@@ -0,0 +1,861 @@
...
+   }
+
+   if (seclevel == 0 && seclevdesc->DeviceOverrideSecurityMinimum ==TRUE ){
+      if (exempt == FALSE){
+         return FALSE;
+      }
+      return TRUE;
+   }
+
+return FALSE;
```
twatteyne: indentation.

openstack/02a-MAClow/IEEE802154_security.h
```diff
@@ -0,0 +1,101 @@
+/**
+\brief Security operations defined by IEEE802.15.4 standard
+
+\author Savio Sciancalepore <savio.sciancalepore@poliba.it>, April 2015.
+\author Giuseppe Piro <giuseppe.piro@poliba.it>,
```
twatteyne: date

openstack/02a-MAClow/IEEE802154_security.h
```diff
@@ -0,0 +1,101 @@
...
+*/
+
+#include "opendefs.h"
+#include "ieee802154_security_driver.h"
+#include "neighbors.h"
+
+extern const struct ieee802154_security_driver IEEE802154_security;
+
+//=========================== define ==========================================
+#define MAXNUMKEYS           MAXNUMNEIGHBORS+1
```
twatteyne: newline missing

openstack/02a-MAClow/ieee802154_security_driver.h
```diff
@@ -0,0 +1,57 @@
...
+#define IEEE802154_SECURITY_KEYIDMODE        0
+#define IEEE802154_SECURITY_KEY_INDEX        0
+#define IEEE802154_SECURITY_TAG_LEN          0
+
+#endif /* L2_SECURITY_ACTIVE */
+
+//=========================== module variables ================================
+
+struct ieee802154_security_driver {
+
```
twatteyne: why these newlines?

openstack/02b-MAChigh/sixtop.c
```diff
@@ -698,18 +705,14 @@ owerror_t sixtop_send_internal(
...
    // transmit with the default TX power
    msg->l1_txPower = TX_POWER;
-   // record the location, in the packet, where the l2 payload starts
-   msg->l2_payload = msg->payload;
    // add a IEEE802.15.4 header
    ieee802154_prependHeader(msg,
                             msg->l2_frameType,
                             payloadIEPresent,
-                            IEEE154_SEC_NO_SECURITY,
+                            msg->l2_securityLevel == IEEE154_ASH_SLF_TYPE_NOSEC ? 0 : 1, // security enabled
```
twatteyne: same as above, please pass `msg->l2_securityLevel` directly

openstack/cross-layers/packetfunctions.c
```diff
@@ -303,6 +303,39 @@ void packetfunctions_tossFooter(OpenQueueEntry_t* pkt, uint8_t header_length) {
    }
 }
 
+//======= packet duplication
+
+void packetfunctions_duplicatePacket(OpenQueueEntry_t* dst, OpenQueueEntry_t* src) {
```
twatteyne: please provide a description of what this function does, and how/when it if used.

openstack/openstack.c
```diff
@@ -2,6 +2,9 @@
 \brief Entry point for accessing the OpenWSN stack.
 
 \author Thomas Watteyne <watteyne@eecs.berkeley.edu>, October 2014.
+
+Modified by:
```
twatteyne: don't put "Modified by", just add your name to the list.

openstack/02a-MAClow/IEEE802154E.c
twatteyne: I don't agree, one can totally build a network with a secret K1 and then running the authentication routine is not just to separate networks.
Yes, please read the ASN into a local variable and use that to authenticate

twatteyne:
@ssciancalepore @malishav 
Very cool stuff, we're almost there! I made a bunch of tiny comments which I believe are super easy to integrate. @malishav , can you confirm that you have been making contributions directly to this PR, maybe throuh forks of @ssciancalepore 's fork?
Congrats to all on the fabulous work!
Thomas :-)

openstack/02a-MAClow/IEEE802154E.c
malishav: Even if K1 is secret, you are relying on leap of faith during join as you are not synced to the network. You are taking ASN from the EB you want to authenticate and then feeding that back into the authentication routing for the EB to pass. We can read the ASN this way *only* during join, all other EBs should be authenticated based on the implicit ASN, not the one present in them. Do you agree?

openstack/02a-MAClow/IEEE802154E.c
malishav: Because it expects bool, not the security level (this corresponds to the security enabled bit in the FCF, not the security level in the security header). We can cast ```ieee154e_vars.ackToSend->l2_securityLevel``` to bool if you think it's more clear?

openstack/02a-MAClow/IEEE802154E.c
twatteyne: +1

openstack/02a-MAClow/IEEE802154E.c
twatteyne: could we not change the signature of `ieee802154_prependHeader()`, if it's always called like this?

malishav:
@twatteyne, yes I was submitting PR's to @ssciancalepore's fork so all the commits are integrated in this PR.

twatteyne:
wonderful!

openstack/02a-MAClow/IEEE802154_security.c
ssciancalepore: @twatteyne at this moment we have only a single key, used to authenticate beacons and encrypting + authenticating all other frames, i.e., data, ack and command frames. 

In the last version of 6TiSCH minimal (v9) I'm reading that "Depending on security policy, these keys could be the same (i.e., K1=K2)". 

Therefore, can we continue to have a single key or it would be a better choice to implement K2 also as a different data string?

malishav: @ssciancalepore can't we solve this by instantiating two keys with different index in init and then setting one index for beacons and the other for data,command, ack frame types?

twatteyne: @ssciancalepore, I would love to have a different K1 and K2 in the code.
Even if one can use K1=K2, IMO that configuration should remain an exception.

ssciancalepore: @twatteyne @malishav yes sure! it's not a problem to have two different keys from the implementation perspective (and it is as Malisa is suggesting). 

My question was only for understanding the configuration we want to provide. It will be fixed in the next pushes.

twatteyne: Wonderful, thanks!!

openstack/02a-MAClow/IEEE802154E.c
malishav: @twatteyne can you take a look if the latest commits address what you had in mind?

twatteyne: for some reason, I get a notification of this comment by mail, but don't
see it in https://github.com/openwsn-berkeley/openwsn-fw/pull/167.
@malisav, can you confirm all suggestions are addressed, or just the one
about ASN?

twatteyne:
@ssciancalepore @malishav ,
I'm very tempted to integrate this code right now, as we need this change to integrate the Golden Device code from @changtengfei.
Can you confirm that you believe the code is ready to roll?
Thomas :-)

ssciancalepore:
@twatteyne @malishav I'm very close to integrate some amendments of the code on my side, based on comments written by Thomas, related to the security processing. They will be pushed by tomorrow morning, at maximum.

Savio

twatteyne:
That's fantastic! Just shout when you are done, and I will merge right away.

openstack/02a-MAClow/IEEE802154E.c
malishav: hm, the new commits outdated the diff where you initially commented so it is hidden now. in any case, I addressed all the suggestions outside of ```IEEE802154_security.[ch]``` files, while @ssciancalepore is working on those!

twatteyne: super!

ssciancalepore:
@malishav the issues that you have highlighted have been fixed, as well as issues indicated by @twatteyne in its review.

On my side, I think that it's all ok. Let me know if there is anything else to do.

Savio

malishav:
OK on my side as well!

twatteyne:
@malishav , @ssciancalepore ,
I'd love to merge, but there appears to be a conflict with the changes done recently. I'm going through the process manually, but I'm not 100% certain I won't break anything. I you want to resolve the conflicts directly in your branch, please feed free.

twatteyne:
I did the merge by hand on my local computer. When a run a simulatoin with `l2_security=0` all is fine. However, when running with `l2_security=1`, Python crashes as soon as I set a mote as DAGroot.

Is is supposed to work in simulation? Was it tested that way?

twatteyne:
Thanks a @ssciancalepore for doing the merge!!

ssciancalepore:
@twatteyne 
I was discussing with @malishav  early this morning, and it seems that there are still problems, deriving by the merge that was done yesterday evening with the official develop. Probably they can be fixed very quickly, because we need just to verify what is changed with latest inclusions.

We wil inform you as soon as the code returns to work perfectly as until yesterday.

twatteyne:
wonderful, thanks!

malishav:
@twatteyne, only tested Python without security, would need to check ```CRYPTO_ENGINE``` integration with Python as it makes part of ```BSP```.

twatteyne:
@malishav, it's perfectly fine if it doesn't support it for now. We should not delay the merge for that. In theory, however, the Python port should just use the SW implementation of CCM*, right?

malishav:
@twatteyne, yes declaration to use ```firmware_crypto_engine``` and some additions to Python build should be all.

twatteyne:
Does the software-only solution implement AES-128 as well?

malishav:
@twatteyne it uses that AES implementation from TI.

twatteyne:
perfect. So shall we declare l2_security and simulation not working for now, and fix later?
"later" should not be too late, as simulation is super helpful to test e.g. packet format of the AUX sec header.

malishav:
OK for me.

twatteyne:
I'll wait for @ssciancalepore to give me the green light to merge.
I created https://openwsn.atlassian.net/browse/FW-401 to track the missing fix for the simulator.

ssciancalepore:
@twatteyne @malishav after the last merge with the official develop we have problems related to the security processing of the ack message.  This is because during IEEE 802.15.4 retrieval process, the code now uses information that should be encrypted (the payload of the ack).

So I'm afraid that the develop now it's incompatible with the security processing and probably a fix need to be done on the develop in order to make both the contributions perfectly integrated.

Can be an idea to integrate this pull request and then create a separate issue for solving the problem on the ack?

Savio

twatteyne:
@ssciancalepore , sure,  we can merge the changes now and fix the issue later. But please understand that we will use L2 sec in one of the tests at the plugtest, so it's important this ACK issue if fixed ASAP.

I created https://openwsn.atlassian.net/browse/FW-402 to track this issue. Could I ask you to fill in the details about the issue directly in there?

twatteyne:
I was able to test that everything works fine without security, i.e. on a 3-note linear network, I can ping and interact over CoAP with the 1st-hop and 2nd-hop mote. I'll merge this one, knowing that we really need https://openwsn.atlassian.net/browse/FW-402 to be fixed in the next couple of days.

malishav:
@twatteyne @changtengfei, problem is that ```ieee802154_retrieveHeader``` now reaches out to the private payload part of the frame in order to parse time correction information from the ACKs and this can not work with security as at that time, frame is still encrypted. We can do ugly work arounds in order to get things back working but IMO this is not the way to go.

 Payload IE parsing should be part of a separate routine invoked after security has passed. @changtengfei, is this something you can take a look at since you introduced the latest changes? Let's discuss this within ```FW-402```.

PR #168, opened 2015-04-10
FW-370. Crypto hardware acceleration for OpenMote-CC2538.

malishav:
This pull request adds the implementation of CCM functions that leverage hardware acceleration specific to CC2538. This crypto engine is used by default for CC2538 motes.


twatteyne:
:+1:

PR #169, opened 2015-04-22
FW-371. Implement functions to read and write RAM on CC2420.

malishav:
This pull request implements functions that read and write CC2420 RAM over SPI. 
Additionally, it modifies ```radio_spiWriteTxFifo()``` in order to generalize it and allow write access to RX FIFO, by accepting the register address argument. Write access to RX FIFO is needed for in-line decryption of CC2420.


twatteyne:
:+1:

PR #170, opened 2015-04-22
FW-372. Crypto hardware acceleration for CC2420 boards.

malishav:
This pull request implements CC2420 hardware-accelerated encryption functions and uses it for TelosB's crypto engine. Stand-alone, 16 byte AES buffer of CC2420 is used for ECB encryption mode. CCM* is implemented by leveraging the in-line functionality of CC2420. CCM* encryption/decryption take place in TX/RX FIFO buffers, respectively. The implementation successfully passes the test suite but has not been fully optimized yet.

### DIFF
--- a/bsp/boards/telosb/SConscript
+++ b/bsp/boards/telosb/SConscript
@@ -29,6 +29,12 @@ source = [
     'board_crypto_engine.c',
 ]
 
+localEnv.Append(
+        CPPPATH =  [
+            os.path.join('#','bsp','chips','cc2420'),
+        ],
+)
+
 board  = localEnv.Object(source=source) + cc2420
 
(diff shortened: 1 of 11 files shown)


twatteyne:
excellent! It looks like this PR contained the same fixes as #169.

malishav:
Thanks. This PR  is dependent on #169 so this branch was rebased on top of it. Is this the way to go or should I rather prepare two independent PRs in future?

twatteyne:
As long as there are no conflicts, that's perfectly fine. The point is to have working code in the repo, and have a (rough) idea of how it got there. 1 PR, 2 PR, 1234 PRs, doesn't matter :-)

PR #172, opened 2015-05-01
FW-368. Fix.

NicolaAccettura:


projects/telosb/SConscript.env
```diff
@@ -24,7 +24,11 @@ MSPVERSION = "1611"
 if   buildEnv['toolchain']=='mspgcc':
    flags = []
    flags += ['-mmcu=msp430f{0}'.format(MSPVERSION)]
-   flags += ['-Os']
+   if buildEnv['debug']==0:
```
twatteyne: Why do you change this?

NicolaAccettura: The -O0 option is needed by Eclipse to debug with gdb when the mspgcc toolchain is used.

twatteyne:
@NicolaAccettura , thanks!!

PR #174, opened 2015-05-07
FW-374. Fix.

changtengfei:
This pull request contains the implementation of sixtop maintenance process and a simple monitoring process according sixtop sublayer draft. Additionally, a sixtop handler is used for identify the entity who call the sixtop reservation protocol. A new subId (new IE)is  used for informing neighbor about the entity of current sixtop handler.


twatteyne:
@changtengfei,
What kind of testing did you do on this code?

changtengfei:
1. I run this code in simulation with three nodes, which topology is linear.
2. mote 1 is set to dagroot, mote 2 is mote 1's child, mote 3 is mote 2's child.
3. using ietf90client to trigger OTF on mote 3, so there will be track through mote 3 to mote 1.
4. the packets can be sent through txrx cell and reserved soft cell. (tx or rx), current pdr on each link is100%
5. Go to topology page, I changed the pdr on the link from mote 3 to mote2 to 0.5
6. after a while, the pdr(numTxAck/numTx) on the cell from mote 3 to mote 2 is lower than a threshold I set. a monitoring process will detect that and call maintenance process to reallocate the cell.
7. the cell between mote 3 to mote 2 has been changed to another one. (a new cell is reserved after remove the old one)
8. the cell between mote 2 to mote 1 is not changed.

changtengfei:
Notice that, if you want to test the code, you should change the NUMSLOTSOFF (in schedule.h) to any value larger than one since the mote 2 at least require two cells, one for Tx parent and one for Rx from child. Or simulation will crash

twatteyne:
@changtengfei , thanks!

PR #175, opened 2015-05-07
always include Python.h first

OlegHahm:
Fixes https://openwsn.atlassian.net/browse/FW-375

The compiler warnings get eliminated according to https://docs.python.org/3.2/c-api/intro.html#include-files by making sure that `Python.h` is always included first.


twatteyne:
@OlegHahm, Wonderful, wonderful!

PR #186, opened 2015-06-07
Golden Image: ROOT and SNIFFER.

changtengfei:
This source code will be used as the golden image presents on plugtest of IETF 93th meeting.


twatteyne:
Per out discussion on the call today, this PR is not meant to be merged into the main repo as-is, at it would break the operation of regular motes.

PR #190, opened 2015-06-27
FW-399. Fixed.

changtengfei:


twatteyne:
## Runnning basic reachability tests

- 3-mote network, linear
- can ping both 1-hop and 2-hop mote
- can interact through CoAP with them

## Testing packet format

- using private IEEE802.15.4e Wireshark dissector, IP-in-IP encapsulation disappeared from DIO

WORKS!

Thanks a million @changtengfei, will merge now.

PR #193, opened 2015-06-28
FW-402. Fix.

ssciancalepore:
Solving problems on ACK decryption and authentication checking.

### DIFF
--- a/bsp/boards/python/supply_obj.c
+++ b/bsp/boards/python/supply_obj.c
@@ -15,6 +15,8 @@
 
 //=========================== public ==========================================
 
+extern int mote_main();
+
 void supply_init(OpenMote* self) {
    
 #ifdef TRACE_ON
--- a/bsp/boards/telosb/eui64.c
+++ b/bsp/boards/telosb/eui64.c
@@ -83,7 +83,7 @@ void eui64_get(uint8_t* addressToWrite) {    // >= 6000us
             *(addressToWrite+0) = 0x14;
             *(addressToWrite+1) = 0x15;
             *(addressToWrite+2) = 0x92;
-            memcpy(addressToWrite+3,id+1,5);
+            memcpy(addressToWrite+3,id+3,5);
          }
       }
    }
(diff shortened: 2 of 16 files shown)


ssciancalepore:
@twatteyne @malishav 
the problem on acknowledgement packets has been solved, and it was due to the fact that the `l2_payload` pointer in the retrieval phase was not set for the ack. Now acknowledgement packets can be exchanged normally also with security enabled.

We have now another little problem: there are packets, whose creator is `COMPONENT_OPENBRIDGE`, that have a total length equal to 123. Therefore, adding 2 bytes of Auxiliary Security Header and 4 bytes of MIC, we exceed the MTU. 
@changtengfei is this correct?

Savio

changtengfei:
@ssciancalepore , if the packet is created by `COMPONENT_OPENBRIDGE`, then it is triggered by the openVisualizer side. The openvisualizer would not send packet to mote automatically except we want to interact with mote. For example when interact with mote through coap or ping the mote, openvisualizer will generate mac payload and sent to mote, then the frame containing the mac payload will be created by `COMPONENT_OPENBRIDGE`. 

Please ignore me if you saw the packet when you interact with the mote. If not, there may some other issue here.

To make sure the size of packet not exceed the MTU, the packets are filtered by following code before sending to mac layer:
https://github.com/openwsn-berkeley/openwsn-fw/blob/develop/openstack/03a-IPHC/openbridge.c#L28

The number in the code should not correct when security is enabled. You can change it accordingly. 
 Let me know if this helps.

ssciancalepore:
@changtengfei thank you for your precious comment!

I see the packet in normal operation, neither during a ping operation nor interacting over CoAP.

Moreover, this packet was generated also before security was merged to the main develop. Can it be possible that the check that you pointed us in `openbridge_triggerData` function need to consider also other bytes added by IEEE 802.15.4 module, apart from security-related parameters?

Savio

twatteyne:
See tests carried out before merge at https://openwsn.atlassian.net/browse/FW-402.

PR #201, opened 2015-07-01
FW-408. Fixed.

changtengfei:


xvilajosana:
@changtengfei  is this also solved in OpenWSN-SW?

as far as I remember, the OpenVisualizer also generates DIOs which may have a MOP option
https://github.com/openwsn-berkeley/openwsn-sw/blob/develop/software/openvisualizer/openvisualizer/RPL/RPL.py#L44

Not sure now if OV is generating the DIOs from the DAGRoot or this are generated by the FW side.
Can you confirm?

changtengfei:
@xvilajosana ,  the DIO was generated by OV before, but now it moved to FW side. But I think what you pointed out need to be fixed, through it will not affect anything.

xvilajosana:
I will open a new issue for SW to clean RPL defines in OV.

twatteyne:
please don't forget to close the issues and delete branch after testin and merging

PR #209, opened 2015-07-04
FW-424. Fix. DAGroot rank still set to 0 when compiling with the option dagroot=1.

NicolaAccettura:


twatteyne:
@NicolaAccettura ,
Great, thanks! I'm testing by running a simulation with a single mote and check the joinPriority field in the EB.
Thomas

twatteyne:
works, thanks!

PR #213, opened 2015-07-13
FW-418. Optional minimal security processing.

malishav:
This PR adds a security attribute `minimal` that can be set to introduce only the minimal necessary security processing without going through table lookups of 15.4 standard. By default, this option is disabled and full security processing is enabled. Enabling security `minimal` means that any key signaling from the frame received over a network will be discarded and the node will use keys known locally (K1 for beacons, K2 for all other frames). 

This PR also omits passing `DATA` frames to the security module before the node is synced, as these frames cannot be decrypted before the first enhanced beacon is received and ASN is known.


twatteyne:
Why don't you use a `#define` for switching the "minimal" mode on/off?

malishav:
Done with ```79b33a```

changtengfei:
@malishav sorry for coming back on this PR so late. Is that hard for you to resolve the conflicts that current PR has, as the Travis build tells.

changtengfei:
Thank you @malishav for solving this! I saw a little changes after the latest commit. Now, it's basically the same as develop branch. I saw there are little changes and make sense. I will merge it.

PR #214, opened 2015-07-15
fix GOLDEN_IMAGE ifdef in openserial

OlegHahm:
Otherwise `type` is not defined in lines 478 and 484.


xvilajosana:
@OlegHahm @changtengfei 

Oleg, It seems this is not needed right? Tengfei can you confirm? If so I will close this PR.

thanks!
X

OlegHahm:
Confirmed.

PR #217, opened 2015-07-18
FW-430. Fix. Disable uinject by default.

NicolaAccettura:


twatteyne:
@NicolaAccettura , I don't quite understand your fix. Why not simply remove `uinject` from `defaultApps`?

NicolaAccettura:
You need to include uinject.h in openapps_dyn.c, since the app is compiled only if included in that file (uinject.h is also included in openudp.c, so at compile time openudp.c will raise an error, because it does not find any source already compiled). To disable uinject by default, the openapps_dyn.c needs only not to call uinject_init.

xvilajosana:
@NicolaAccettura 

I see you are comenting cexample_init. Is this uinject? 

X

NicolaAccettura:
@xvilajosana,

let's start from writing down something very well known, just to agree on the desired behavior.

Openapps.c will be used in fact only in simulation mode, and there is no possibility to dynify it in such a mode. Cexample and cstorm are added to the default apps (in the sconscript file), although they will be really enabled only if initialized in openapps.c. Cstorm is initialized in simulation mode because it is initialized in openapps.c. Cexample was not initialized in openapps.c. I've added the commented line for some sort of organization. Do we want cexample to be a default app in simulation mode? If yes, the commented cexample_init() should be uncommented. And in case, I would delete the following lines in the sconscript file:

if localEnv['board']=='python':
     defaultApps  += [
         'cexample',
         'cstorm',
     ]

since in practice they are useless (openapps.c is not dinyfied in simulation mode).
Cexample is not related to uinject, we agree on that. By default, uinject is not enabled in simulation mode, since cexample_init is also commented out.
The behavior with real motes is different, to initialize apps, the dynified openapps_dyn.c is used. In that case, openapps.c is not used. We need to include the uinject.h file in the openapps_dyn.c, while not initializing it (the fix in this PR allow this distinction between included apps and enabled apps).
Hope to have been clear.
Nicola

twatteyne:
Great, thanks!

PR #218, opened 2015-07-18
FW-431. Fix.

NicolaAccettura:
At the beginning the tx attempts left are now TXRETRIES plus one.


twatteyne:
great, thanks for catching this.

PR #220, opened 2015-07-18
FW-433. Big endian format for ASN part of the security nonce.

malishav:
This PR creates security nonce using big endian format of ASN.


malishav:
@changtengfei, I just rebased this PR. Should be ready to fly, take a look. Had to use a special packetfunction because ASN is 5 bytes long.

PR #222, opened 2015-07-30
FW-437. Fixed.

changtengfei:


Erfanma:
Hi @changtengfei 
Why don't you use "MAXACTIVESLOTS"?? Because it is possible to have MAXACTIVESLOTS < schedule_getFrameLength().
Using "schedule_getFrameLength()" it's possible to add cells in OFF slotoffsets. Doesn't make problem??

changtengfei:
@Erfanma , the slotoffset should be the number from 0 to slotframeLength-1, not from 0 to MAXACTIVESLOTS. MAXACTIVESLOTS is the buffer size of schedule.

Erfanma:
@changtengfei,  your right. Thanks!

changtengfei:
:-)

PR #223, opened 2015-07-31
FW-439. Fix.

twatteyne:


twatteyne:
Thanks a million to @elodimel and @ddujovne for helping fix this!

PR #227, opened 2015-08-04
FW-443. Fix. Spourious EBs.

NicolaAccettura:


openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -2045,8 +2067,10 @@ void notif_sendDone(OpenQueueEntry_t* packetSent, owerror_t error) {
 void notif_receive(OpenQueueEntry_t* packetReceived) {
    // record the current ASN
    memcpy(&packetReceived->l2_asn, &ieee154e_vars.asn, sizeof(asn_t));
-   // indicate reception to the schedule, to keep statistics
```
twatteyne: I believe `isSpouriousEB` can never be `TRUE` in an RX cell, so I would revert the changes to `notif_receive`.

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -782,18 +781,12 @@ port_INLINE bool ieee154e_processIEs(OpenQueueEntry_t* pkt, uint16_t* lenIE) {
...
-            ieee154e_vars.freq = 11 + (asnOffset + channelOffset)%16 
-            */
-            for (i=0;i<16;i++){
-                if ((ieee154e_vars.freq - 11)==ieee154e_vars.chTemplate[i]){
-                    break;
-                }
+            if (schedule_syncSlotOffset(ieee154e_vars.slotOffset)==TRUE) {
+               ieee154e_vars.nextActiveSlotOffset = schedule_getNextActiveSlotOffset();
+            } else {
+               ieee154e_vars.isSpouriousEB = TRUE;
```
twatteyne: The name `isSpouriousEB` is a bit misleading. Is there not a more comprehensive name?

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -874,13 +870,32 @@ port_INLINE void activity_ti1ORri1() {
...
-      //start outputing serial
-      openserial_startOutput();
+      // check if I can send an EB
+      ieee154e_vars.dataToSend = openqueue_macGetEBPacket();
+      if (ieee154e_vars.dataToSend != NULL) {
+         ieee154e_vars.isSpouriousEB = TRUE;
+         // change state
+         changeState(S_TXDATAOFFSET);
+         // change owner
+         ieee154e_vars.dataToSend->owner = COMPONENT_IEEE802154E;
```
twatteyne: Please move the following lines to a function which gets called here, and in the place to copy-pasted this code from.

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -997,7 +1012,11 @@ port_INLINE void activity_ti2() {
    packetfunctions_reserveFooterSize(&ieee154e_vars.localCopyForTransmission, 2);
    
    // calculate the frequency to transmit on
-   ieee154e_vars.freq = calculateFrequency(schedule_getChannelOffset()); 
+   if (ieee154e_vars.isSpouriousEB == TRUE) {
+      ieee154e_vars.freq = calculateFrequency(SCHEDULE_MINIMAL_6TISCH_CHANNELOFFSET);
```
twatteyne: This is a great idea. Please:
- declare a new variable `EB_CHANNELOFFSET`, set to `15`
- return an error when a module tries to add a cell on channelOffset equal to `EB_CHANNELOFFSET`.

This will avoid collisions between EB and other dedicated traffic.

openstack/02b-MAChigh/schedule.c
```diff
@@ -521,6 +529,49 @@ slotOffset_t schedule_getNextActiveSlotOffset() {
 }
 
 /**
+\brief return slotOffset of closest active slot
```
twatteyne: In the comments, add:
- when this function is called and by which module
- what does "closest" mean? Only in the future, or also in the future AND past.

NicolaAccettura: This is called when receiving the first EB by activity_synchronize_endOfFrame. It return the closest future active slot.

openstack/02a-MAClow/IEEE802154E.c
NicolaAccettura: The channel offset used for sending EB is SCHEDULE_MINIMAL_6TISCH_CHANNELOFFSET, since EBs have to be sent also in the shared cell of minimal.
It has been added a check in schedule_addActiveSlot: if the cell going to be added is on SCHEDULE_MINIMAL_6TISCH_CHANNELOFFSET and set as TX or RX, it will not be added.

openstack/02a-MAClow/IEEE802154E.c
NicolaAccettura: That flag is set TRUE at the beginning of slot when an unscheduled EB will be sent, or when receiving the first EB (look ieee154e_processIE). It is used to avoid having statistics showed for cells that will usually be OFF.

openstack/02a-MAClow/IEEE802154E.c
NicolaAccettura: Agree, changed to isUnscheduledEB.

openstack/02a-MAClow/IEEE802154E.c
NicolaAccettura: Done.

changtengfei:

PR #229, opened 2015-08-17
FW-445. Unscheduled EBs and slotskip optimization. Fix.

NicolaAccettura:


changtengfei:

PR #230, opened 2015-10-01
FW-449. Fixing overflow in DAGrank calculation

koalo:
In the calculation of the tentativeDAGrank in
neighbors_updateMyDAGrankAndNeighborPreference an overflow was possible.
This results in occasional choosing a preferred parent with MAXDAGRANK.

### DIFF
--- a/openstack/02b-MAChigh/neighbors.c
+++ b/openstack/02b-MAChigh/neighbors.c
@@ -543,7 +543,8 @@ void neighbors_updateMyDAGrankAndNeighborPreference() {
             rankIncrease = (uint16_t)(rankIncreaseIntermediary >> 10);
          }
          
-         tentativeDAGrank = neighbors_vars.neighbors[i].DAGrank+rankIncrease;
+         // cast the uint16_t summands to avoid an overflow if DAGrank == 0xFFFF (MAXDAGRANK)
+         tentativeDAGrank = (uint32_t)neighbors_vars.neighbors[i].DAGrank + (uint32_t)rankIncrease;
          if ( tentativeDAGrank<neighbors_vars.myDAGrank &&
               tentativeDAGrank<MAXDAGRANK) {
             // found better parent, lower my DAGrank


openstack/02b-MAChigh/neighbors.c
```diff
@@ -543,7 +543,8 @@ void neighbors_updateMyDAGrankAndNeighborPreference() {
             rankIncrease = (uint16_t)(rankIncreaseIntermediary >> 10);
          }
          
-         tentativeDAGrank = neighbors_vars.neighbors[i].DAGrank+rankIncrease;
+         tentativeDAGrank = neighbors_vars.neighbors[i].DAGrank;
+         tentativeDAGrank += rankIncrease;
```
changtengfei: @koalo , could you explain how your changes works? It seems they are exactly the same?

koalo: tentativeDAGrank is uint32_t, while rankIncrease and DAGrank are uint16_t. Since MAXDAGRANK is 0xFFFF, the sum might be larger than uint16_t so the result might be smaller than 0xFFFF. When adding the summands one by one the overflow does not occur. An alternative would be 
tentativeDAGrank = neighbors_vars.neighbors[i].DAGrank+(uint32_t)rankIncrease;
Do you prefer that one?

changtengfei: @koalo yours is good! I just didn't see the type of the two variables. 
Congratulations on finding this! I will merge this code after the build passed :)

koalo: I will add a short comment to the patch to avoid that someone "unfixes" this later. (I think travis can handle another build...)

koalo:
There is something strange with the Travis CI build... Any idea?

changtengfei:
Yes, I noticed that. It seems something wrong to download the texlive-binaries from the server.

changtengfei:
@koalo , could I ask you to change the line to the following which should be more easy to understand? Thanks!
`tentativeDAGrank = (uint32_t)neighbors_vars.neighbors[i].DAGrank+(uint32_t)rankIncrease;`

koalo:
Yes of course! The patch is updated.

changtengfei:
This is perfect! Thanks a lot!

PR #231, opened 2015-10-01
SW-212. Fix baud rate for IoT-LAB M3 nodes

koalo:
Fixing the baud rate for the M3 nodes in the IoT-LAB. Has to be 500 kBaud.


changtengfei:
@koalo I saw in the repository, the baud rate is changed to 500000. Will you have extra work on pull request? (I am considering you may need modify some code in openwsn-fw. If you are going to push your work here, I will wait for you. If not, I will close this.) Thanks!

koalo:
Yes, this patch is no longer necessary since the fix was also contained in #237.

PR #233, opened 2015-12-09
Avoiding conditional directives that break statements

RomeroMalaquias:


twatteyne:
Thanks for your contribution @RomeroMalaquias. Please note that we are not actively maintaining the K20 port, but I'm happy to merge your contributions.
Thomas

PR #240, opened 2016-03-10
FW-435. Fixed.

changtengfei:


openstack/cross-layers/idmanager.c
```diff
@@ -194,6 +206,7 @@ void idmanager_triggerAboutRoot() {
    switch (input_buffer[0]) {
      case ACTION_YES:
         idmanager_setIsDAGroot(TRUE);
+        idmanager_vars.slotSkip = FALSE;
         break;
      case ACTION_NO:
         idmanager_setIsDAGroot(FALSE);
```
xvilajosana: @changtengfei what happens if I am dagroot, and someone toggles me to become a node in the network? Should we say here idmanager_vars.slotSkip=TRUE; ? Note that has been set to false when the node turned to dagroot before.

changtengfei: @xvilajosana , yes, it should be set to TRUE. I will fix this in a second.

PR #245, opened 2016-04-18
FW-466. implemented light sensor functions for TelosB platform

pedrohenriquegomes:
I created all functions to implement the sensor on TelosB. Only light sensors are working at this time, but it will be easy to extend the implementation to other sensor later.


changtengfei:
@pedrohenriquegomes , That's great! Thanks a lot for implementing this! I will merge this now!

PR #247, opened 2016-05-01
removing k20 and xpressohack

xvilajosana:
getting rid of old k20 and xpressohack. Please have a safety check.


changtengfei:
Tested with 4 nodes linear network. All ping command rearched. firefox Copper plugin work fine.

PR #249, opened 2016-05-04
FW-484. add comment after defining NUMSLOTSOFF.

changtengfei:


Erfanma:
Hi Tengfei,
I think you should correct the comment, it should be "add" instead of "added"! :-)

changtengfei:
Erfan, thanks for indicating that! I will correct it with another branch I am working on.

PR #250, opened 2016-05-09
FW-485 modifying bootloader to load ihex instead of elf. Updating dat…

xvilajosana:
this modifyies the bootloader command for openmote so we load intel ihex file instead of the elf. It also pushes the bootloader speed to 400000 instead of 115200

use pip install intelhex in case intelhex is not present


twatteyne:
can we put `intelhex` in the `requirements.txt` file and instruct people to run `pip install -r requirements.txt`?

xvilajosana:
added intelhex to requirements directly

twatteyne:
thanks!

PR #251, opened 2016-05-09
fixing some compilation errors in armgcc

xvilajosana:
compiling with lots of warnings. Needs a second round.

MAC no synchronizing with 2 motes,
14:05:50 ERROR 9c86 [IEEE802154E] invalid packet from radio, length 249 (code location 0)

to check wit sniffer and see if EBs are sent correctly.
to check MAC layer timing.


xvilajosana:
do not merge this, sorry was an error.

PR #257, opened 2016-05-18
Fw 480

JRubis:
FW-480 Slotskip optimization and deSyncTimeout. Fixed and verified timeout is properly working with skipslot enable or disabled.


changtengfei:
@JRubis , This is a great fix! Thanks a lot! I will merge this.

PR #258, opened 2016-05-18
FW-468. Fixed the way counter is written in the uinject app. 

pedrohenriquegomes:
And fixed 'uinject.py' script to read both ASN and the counter.


changtengfei:
Thanks for fixing this @pedrohenriquegomes  :)

PR #259, opened 2016-05-23
ADM-62 adding ide functionality to scons builder

xvilajosana:
this adds a new functionality to scons builder system. By specifying the [ide] e.g., ide=qtcreator the IDE is configured automatically according to the board we are compiling. for example
scons board=OpenMote-CC2538 toolchain=armgcc ide=qtcreator oos_openwsn. 

this runs qtcreator (must be in the path) configured for the openmote platform. We can run it for gina, telosb, etc...

qtcreator is opensource and can be downloaded from here 
http://www.qt.io/download-open-source/

or installed in linux by
sudo apt-get install qtcreator

no need to register, just skip and have fun. it is a lightweight and fast ide with very nice functionalities.
It is tested in windows and linux. Note that qtcreator must be in the path.

### DIFF
--- a/SConscript
+++ b/SConscript
@@ -5,6 +5,7 @@ import subprocess
 import platform
 import distutils.sysconfig
 import sconsUtils
+from tools import qtcreator as q
 
 Import('env')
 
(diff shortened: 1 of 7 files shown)


changtengfei:
Tests:
1. python compilation passed
2. simulating network is able to setup and the mote can be reached by ping command.
3. coap copper app works
4. compiling openmotecc2538, gina. telosb, iotlab_m3, openmotestm with ide=qtcreator options works.

twatteyne:
Wow, totally cool. @changtengfei, could you please delete the branch once merged (check the button "Delete branch" right on this page, no need to do this by hand in Git)

changtengfei:
Got it! :)

twatteyne:
Thanks!

PR #263, opened 2016-05-24
FW-497. For CoAP, use a NON type response for a NON type request.

kbonset:
* Use an ACK response only for a CON request.


changtengfei:
@kbonset ,thanks for fixing this!

PR #305, opened 2016-10-12
FW-490. Add support for MSP430X processors

MrD2:
- Changed the compiler used from msp430-gcc to msp430-elf-gcc
- Added a board and project folder where there are files and macros for compiler to use extended compilation
- Fix issues with compiler definitions for other boards that use mspgcc


twatteyne:
@MrD2 , thanks for the contribution. I'll you discuss with @changtengfei about resolving the checks which are currently failing.

changtengfei:
@MrD2 is there a easy way to install the msp430-elf-gcc, like sudo apt-get install msp430-elf-gcc? To make the travis builder pass this new compiler, it's required to install it on the travis builder machine. It's configured in this file.
https://github.com/openwsn-berkeley/openwsn-fw/blob/develop/.travis.yml

MrD2:
Short answer. No.

Not so short answer, i already implemented the new toolchain on the openvm repo.
I can try and update the travis.yml file to make the same installation/configuration, but i never worked with that.

changtengfei:
@MrD2 have you got a chance to make Travis installing the msp-elf-gcc?

MrD2:
Not yet, 

I am compiling the code using the OpenWSN VM.... has i have problems with the shared folders it's a pain to use the git on cmd line :P

Also i remenbered that the modifications i did somewhat disrupt cross-compiler compatibility because if the interruptions definitions are set on the interrupt_def.h (or something) it throws an error because it dosent accept MAKROS on interrupts. I still have to check this and find a work around.

Just out of curiosity, can i install stuff on travis like a .sh script on the cmd? and use the /opt path or other?

changtengfei:
@MrD2 certainly you can install things on travis. The .travis.yml is the file you are going to play with. 
Also I believe accessing a path shouldn't be a problem. You can refer to here to see what is capable to do with yml script: https://docs.travis-ci.com/user/installing-dependencies/

MrD2:
I am currently importing my code here.... you can check if you want on my personal fork. 
I still have to fix things as i noticed some bsp sources were changed and at the same time i will import and update MSP430X compatibility, update FreeRTOS and upload my own project as a MSP430X example

Best regards

MrD2:
Also, if someone reads this before i have a change to update the files @changtengfei could you resume the changes made to board_crypto_engine.c, bsp_timer.c, and radiotimer.c fery quick?

changtengfei:
@MrD2 , since we are using sctimer to implement all timers used in openwsn, there is no bsp_timer and radiotimer anymore.

PR #314, opened 2016-11-16
No radio retune needed for ACK send/receive

mikehaben:
With 2 nodes side-by-side on the bench (so good radio reception), I was seeing ACKs occasionally not being picked up when frequency-hopping, but not when running single-frequency.  With the radio-retune lines of code deleted from activity_ti6() and activity_ri6(), 100% of ACKs are received.

### DIFF
--- a/openstack/02a-MAClow/IEEE802154E.c
+++ b/openstack/02a-MAClow/IEEE802154E.c
@@ -1169,12 +1169,6 @@ port_INLINE void activity_ti6() {
    // change state
    changeState(S_RXACKPREPARE);
    
-   // calculate the frequency to transmit on
-   ieee154e_vars.freq = calculateFrequency(schedule_getChannelOffset()); 
-   
-   // configure the radio for that frequency
-   radio_setFrequency(ieee154e_vars.freq);
-   
    // enable the radio in Rx mode. The radio is not actively listening yet.
    radio_rxEnable();
    //caputre init of radio for duty cycle calculation
@@ -1664,12 +1658,6 @@ port_INLINE void activity_ri6() {
     // space for 2-byte CRC
    packetfunctions_reserveFooterSize(ieee154e_vars.ackToSend,2);
   
-    // calculate the frequency to transmit on
-   ieee154e_vars.freq = calculateFrequency(schedule_getChannelOffset()); 
-   
-   // configure the radio for that frequency
-   radio_setFrequency(ieee154e_vars.freq);
-   
    // load the packet in the radio's Tx buffer
    radio_loadPacket(ieee154e_vars.ackToSend->payload,
                     ieee154e_vars.ackToSend->length);
--- a/SConscript
+++ b/SConscript
@@ -193,7 +193,7 @@ elif env['toolchain']=='iar-proj':
     
 elif env['toolchain']=='armgcc':
     
-    if env['board'] not in ['OpenMote-CC2538','iot-lab_M3','iot-lab_A8-M3','openmotestm']:
+    if env['board'] not in ['OpenMote-CC2538','iot-lab_M3','iot-lab_A8-M3','openmotestm', 'samr21_xpro']:
         raise SystemError('toolchain {0} can not be used for board {1}'.format(env['toolchain'],env['board']))
     
     if   env['board']=='OpenMote-CC2538':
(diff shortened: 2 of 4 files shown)


changtengfei:
@mikehaben , I didn't get the problem you are trying to descript. Is there an issue related to this PR?

openstack/02a-MAClow/IEEE802154E.c
```diff
@@ -1169,12 +1169,6 @@ port_INLINE void activity_ti6() {
    // change state
    changeState(S_RXACKPREPARE);
    
-   // calculate the frequency to transmit on
-   ieee154e_vars.freq = calculateFrequency(schedule_getChannelOffset()); 
-   
-   // configure the radio for that frequency
-   radio_setFrequency(ieee154e_vars.freq);
```
twatteyne: Hmm, this completely disables channel hopping. Is there a reason?

twatteyne:
I like the fact that you add SCons support for the SAM R21 very much, but you disable channel hopping which is a big no-no

twatteyne:
@mikehaben,
What you are seeing is that some frequencies are less good than others because of interference or multi-path fading. This is exactly why channel hopping is a good idea: if a frequency doesn't work, the re-transmit will be on a different one. So this is why we want it :-)

mikehaben:
Sorry, I must be misunderstanding something - I thought that the 
radio-retune happens at the start of the timeslot, in either 
activity_ti2() if about to transmit, or activity_ri2() if about to 
receive, and that the radio doesn't need to be retuned again within the 
same timeslot to transmit or receive the ACK.  Or is there the 
possibility that it's going to handle an ACK related to a different 
packet / node?

I'd be very surprised if this was genuinely an RF issue - the two nodes 
are ~20cm apart, and my RTL-SDR +SDRsharp "spectrum analyser" isn't 
showing any other transmissions in the band.


Mike Haben
Senior Embedded Systems Engineer | ASH Wireless Electronics Ltd.

Email: mh@ashwireless.com
Phone: +44 (0)23 8055 1044
Website: www.ashwireless.com

ASH | Creative Wireless Electronics

ASH Wireless Electronics Ltd.
Leornain House | Itchen Business Park
Kent Road | Southampton
SO17 2LJ | United Kingdom

ASH Wireless Electronics is a limited company registered in England and 
Wales. | Company Registration No. 4148550 | Registered Offices: 5 
Shaftesbury Avenue, Southampton, SO17 1SB, United Kingdom. This message 
contains confidential information and is intended only for the named 
Recipients. If you are not the intended recipient you should not 
disseminate, distribute or copy this e-mail. Please notify the sender 
immediately by e-mail if you have received this e-mail by mistake and 
delete this e-mail from your system. E-mail transmission cannot be 
guaranteed to be secure or error-free as information could be 
intercepted, corrupted, lost, destroyed, arrive late or incomplete, or 
contain viruses. ASH Wireless Electronics Ltd therefore does not accept 
liability for any errors or omissions in the contents of this message, 
which arise as a result of e-mail transmission. If verification is 
required please request a hard-copy version.The opinions, advice or 
judgements within this email are not necessarily the views shared by ASH 
Wireless Electronics Ltd.



------ Original Message ------

twatteyne:
Oh I see! I hadn't expanded the diff in GitHub and didn't see that you were talking about the setting of the frequency before sending/receiving the ACK! Yep, it looks overkill to reset the frequency as the radio isn't turned off between DATA and ACK, But are we sure none of the radios need re-setting the frequency when turning around between RX and TX.

@changtengfei, any input?

changtengfei:
All radios I know currently wouldn't change the frequency after switching between Tx and Rx, though, I am not sure this is true for all radio existed. If yes, indeed, we needn't to set the frequency again.

twatteyne:
Ok, thanks. Could it mean that we have some borderline timing problem though? I don't understand why configuring the frequency would mess up reception though?

mikehaben:
I'm using a sub-GHz radio (Atmel AT86RF215, same as Jonathan Muñoz), 
could be specific to this device.

changtengfei:
@mikehaben according to the description above, to make sure all radio chips works, re-setting the frequency again when sending back ACK is required. But I would like to keep the part the integrate SAM R21 into Scons. Could you reverse the changes in IEEE802154E.c? I will merge your PR then. Thanks!

changtengfei:

changtengfei:
The scons support for samr21_xpro is already merged into repo. This PR will be closed.

PR #319, opened 2016-12-08
FW-589. Mercator project crashes when receiving wrong HDLC frame

keomabrun:


twatteyne:
Is that what caused the misterious problems in IoT-lab?

keomabrun:
I am not sure of that no. But the code should no crash if it occurs anyway.

twatteyne:
ok

kernel/openos/scheduler.c
```diff
@@ -52,9 +52,7 @@ void scheduler_start() {
          pThisTask->next          = NULL;
          scheduler_dbg.numTasksCur--;
       }
-      debugpins_task_clr();
```
changtengfei: The debugpins functions here shouldn't be deleted as it's used for other board.

kernel/openos/scheduler.c
```diff
@@ -52,9 +52,7 @@ void scheduler_start() {
          pThisTask->next          = NULL;
          scheduler_dbg.numTasksCur--;
       }
-      debugpins_task_clr();
       board_sleep();
-      debugpins_task_set();                      // IAR should halt here if nothing to do
```
changtengfei: Same as previous one.

changtengfei:
@keomabrun cool! I left two tiny comments. After they are solved, I will merge the code..

PR #329, opened 2016-12-16
FW-606. Added missing ISR lock in scheduler.

medarion:


changtengfei:
@medarion Thanks for the input! I checked the Travis status and found you need to declare the interruption before using the DISABLE/ENABLE interrupt function. 
Add this code at the beginning of the function 'scheduler_start' : `INTERRUPT_DECLARATION` should solve the Travis failed status.

medarion:
Thanks. It checks out now.

PR #330, opened 2016-12-20
FW-609. Schedule new tasks after old ones of the same priority

medarion:


changtengfei:
Merged! Thanks a lot! @medarion

PR #331, opened 2016-12-23
FW-611. Support for OpenMote-CC2538 Rev.A1

peretuset:
This pull request adds many goodies to the OpenMote-CC2538 platform as described next:
1) Ability to flash multiple boards at once to save time. It is done using bootload=/dev/ttyUSBX (which expands to all) or bootload=/dev/ttyUSB[0-1] (which expands to a range). Passing a single valid argument works as it did before.
2) Add support for the OpenMote-CC2538 rev.A1 which comes with the CC2538SF23 (256 KB Flash) instead of the CC2538SF53 (512 KB Flash). To support it a revision parameter has been added to SCons. By default it reverts to the CC2538SF53, but passing revision=A1 changes the linker script accordingly.
3) Renamed the OpenMote-CC2538 platform to openmote-cc2538 as requested a while ago. Now compiling is done with platform=openmote-cc2538 instead of platform=OpenMote-CC2538.
4) Flashing is now automatic with the the new OpenUSB rev.A1 board. The OpenUSB takes care of putting the CC2538 in bootload mode and, after that, it flashes the code using the cc2538-bsl.py script. Also, the flashing speed has been increased to 500000 to reduce the flashing time.
5) Updated the includes in the SCons build system to avoid having to modify the CC2538 Foundation Firmware provided by TI.  
All changes have been tested and should work out of the box.


peretuset:
@twatteyne @changtengfei Please take a look when you can and see if this can be merged in.

twatteyne:
@changtengfei , can I ask you to review and merge?

SConscript
```diff
@@ -501,7 +543,7 @@ class OpenMoteCC2538_bootloadThread(threading.Thread):
     def run(self):
         print 'starting bootloading on {0}'.format(self.comPort)
         subprocess.call(
-            'python '+os.path.join('bootloader','OpenMote-CC2538','cc2538-bsl.py')+' -e -w -b 400000 -p {0} {1}'.format(self.comPort,self.hexFile),
+            'python '+os.path.join('bootloader','openmote-cc2538','cc2538-bsl.py')+' -e --bootloader-invert-lines -w -b 500000 -p {0} {1}'.format(self.comPort,self.hexFile),
```
changtengfei: @peretuset I remember there is an issue after changing the baudrate to 500000, is it now solved? Currently I havn't openmote-cc2538 on my side, I will test after I back from vacation.

changtengfei:
@peretuset, I have reviewed your code and find you include header file with colon("header_file") and angle bracket(<header_file>). Do you have any concern on doing so?

peretuset:
@changtengfei, this is compiler dependent, but using colons prioritizes headers in the current working directory over system headers, i.e. the compiler looks first in the current directory for the included files.
However, brackets <> are typically used for system headers, so I am just trying to keep it how it should be to indicate those included headers are system headers. Is that a problem?

changtengfei:
No problem, @peretuset . I will test this code with CC2538 Mote today and merge it if everything goes well :-)

changtengfei:
@peretuset  the baud-rate of 500000 still doesn't work on my side. I am working on Windows. 
I have tested the point 1,3 and 5 you pointed out and it works fine with 400000 baudrate.
For point 2 and 4, I can't test them since we don't have the new version board.

peretuset:
@changtengfei this is probably because you are still using the old hardware. Anyway, I do not see any problem to reduce it to 400000 to make it backwards compatible. I will change and update the PR.

changtengfei:
Merged! @peretuset , I think we don't have the new version of openmote-cc2538. Probably we could get some in the future. Thanks a lot for doing this support!

PR #332, opened 2017-01-10
FW-446. Fixed timer period computation when synchronizing for the fir…

medarion:
…st time.


changtengfei:
Merged. @medarion , thanks a lot for the fixing this issue! Cool!

PR #339, opened 2017-01-26
FW-583. TelosB mote is on for 1ms too long because radio_rxEnable() is called in activity_ri2 rather than activity_ri3.

TimothyClaeys:
TelosB mote is on for 1ms too long because radio_rxEnable() is called
in activity_ri2 rather than activity_ri3.


changtengfei:
@TimothyClaeys Thanks for fixing this! Cool!

PR #342, opened 2017-03-07
FW-628. Time correction is not standard compliant.

changtengfei:


twatteyne:
don't pull.

sieben:
I've pushed a commit to wireshark upstream to enhance the handling of time correction dissection. In particular it knows produces a numeric interpretation of the time shift instead of just returning the raw bytes.  https://code.wireshark.org/review/#/c/20547/2/epan/dissectors/packet-ieee802154.c

sieben:
Is there any tricks/tips to generate lots of ACK? I've tried to increase the traffic towards the nodes but it didn't increased the amount of ACK I'm seeing.

changtengfei:
@sieben are you able to see the traffic packet increased? If the packet associated to the traffic requiring ACK, there should have lots of ACK sent out by increasing the traffic, as long as they are synchronized.

PR #348, opened 2017-05-24
Functions for CoAP options

otibsa:
This PR refactors the opencoap module and applications.

It provides a way for the applications to extract CoAP options from incoming requests and to set options in responses without worrying about maintaining the correct order of the options (increasing option type) or their encoding (option delta, extended option fields) or adding the payload marker.
This is mainly done through the functions opencoap_getOption(...) and opencoap_setOption(...).

Please let me know if this is useful and/or if I should change something.


changtengfei:
@otibsa thanks for creating this PR. Have you encountered any problem with the code before? Could you explain the issue you had? Thanks!

otibsa:
I am working on a coap application for which I need the coap block option for blockwise transfer of large requests. Some of my commits where only focused on cleaning up the way coap options are handled, so I created a branch for that. 
I also have a branch for the block option, and I will create another pull request, once it works fine.

chris-ho:
@changtengfei @otibsa i'm supporting this PR as well its very useful if you want to deal with multiple options for my application i'm using e.g observer and query

changtengfei:

PR #350, opened 2017-05-31
CoAP blockwise transfer

otibsa:
This PR adds support for blockwise transfer ([RFC7959](https://tools.ietf.org/html/rfc7959)) to opencoap.

Both Block2 (for large responses) and Block1 (for large requests) options are implemented.
This means, that an opencoap application can generate responses that don't have to reuse `msg->payload` from the original request, but instead point `msg->payload` to some memory area that won't be changed after the first request with a Block2 option is received (i.e. not the stack). These responses also don't have to be limited by 127 Bytes but can be larger.

Please let me know if this is useful and/or if I should change something.


changtengfei:

PR #351, opened 2017-06-08
FW-649. uses a larger set of outputbuffer to avoid dropping some data…

ftheoleyre:
by default 5 buffers of 256 bytes to store the data before its transmission through the serial line (avoids to drop data)


changtengfei:

changtengfei:
The SERIAL_OUTPUT_BUFFER_SIZE is changed to 1024 and use a modulo operator of 0x3FF.
Hence the output buffer is increased.

PR #369, opened 2017-07-19
FW-681 fixed handling of coap options

chris-ho:


malishav:
Hi @chris-ho,

Thanks a bunch for providing this PR. I see that there are a lot of .ewp and .ewt files committed, could you remove those changes from the PR in order to have a clean merge. You can keep them on your local branch, just don't push it on the PR branch. I will leave some comments on the code inline.

Mališa

openapps/cinfo/cinfo.c
```diff
@@ -106,13 +106,14 @@ owerror_t cinfo_receive(
...
-         msg->payload[sizeof(infoStackName)-1+5-4] = '.';
-         msg->payload[sizeof(infoStackName)-1+5-3] = '0'+OPENWSN_VERSION_MINOR;
-         msg->payload[sizeof(infoStackName)-1+5-2] = '.';
-         msg->payload[sizeof(infoStackName)-1+5-1] = '0'+OPENWSN_VERSION_PATCH;
+         msg->payload[sizeof(infoStackName)-1+6-6] = '0'+OPENWSN_VERSION_MAJOR;
+         msg->payload[sizeof(infoStackName)-1+6-5] = '.';
+         msg->payload[sizeof(infoStackName)-1+6-4] = '0'+OPENWSN_VERSION_MINOR / 10;
+         msg->payload[sizeof(infoStackName)-1+6-3] = '0'+OPENWSN_VERSION_MINOR % 10;
+         msg->payload[sizeof(infoStackName)-1+6-2] = '.';
+         msg->payload[sizeof(infoStackName)-1+6-1] = '0'+OPENWSN_VERSION_PATCH;
```
malishav: Why are these changes needed?

openapps/opencoap/opencoap.c
```diff
@@ -169,15 +171,31 @@ void opencoap_receive(OpenQueueEntry_t* msg) {
 
    // process handled options
    //== Stateless Proxy option
-   statelessProxy = opencoap_find_option(coap_incomingOptions, coap_incomingOptionsLen, COAP_OPTION_NUM_STATELESSPROXY);
+   option_count = opencoap_find_option(coap_incomingOptions, coap_incomingOptionsLen, COAP_OPTION_NUM_STATELESSPROXY, &option_index);
+   if(option_count >= 1)
```
malishav: We typically use the syntax:

if(condition) { 

} else {

}

with braces on the same line as the if/else condition. Same for other loops. Could you update your PR to reflect this?

openapps/opencoap/opencoap.c
```diff
@@ -846,19 +873,23 @@ owerror_t opencoap_options_encode(
...
+uint8_t opencoap_find_option(coap_option_iht* array, uint8_t arrayLen, coap_option_t option, uint8_t* startIndex) {
+   uint8_t i, j;
 
-    for (i = 0; i < arrayLen; i++) {
-        if (array[i].type == option) {
-            return &array[i];
-        }
-    }
-    return NULL;
+   for (i=0; i<MAX_COAP_OPTIONS; i++) {
```
malishav: Why don't you use `arrayLen` instead of `MAX_COAP_OPTIONS` define?

openapps/opencoap/opencoap.c
```diff
@@ -958,20 +989,38 @@ void opencoap_handle_proxy_scheme(OpenQueueEntry_t *msg,
...
     coap_option_iht *proxyScheme;
     const uint8_t proxySchemeCoap[] = "coap";
     const uint8_t uriHost6tisch[] = "6tisch.arpa";
     open_addr_t JRCaddress;
 
     // verify that Proxy Scheme is set to coap
-    proxyScheme = opencoap_find_option(incomingOptions, incomingOptionsLen, COAP_OPTION_NUM_PROXYSCHEME);
+    option_count = opencoap_find_option(incomingOptions, incomingOptionsLen, COAP_OPTION_NUM_PROXYSCHEME, &option_index);
+    if(option_count >= 1)
+    {
```
malishav: same comment about `if` syntax applies throughout the code.

openapps/opencoap/opencoap.c
```diff
@@ -846,19 +873,23 @@ owerror_t opencoap_options_encode(
...
 }
 
-coap_option_iht* opencoap_find_option(coap_option_iht* array, uint8_t arrayLen, coap_option_t option) {
-    uint8_t i;
-
-    if (array == NULL || arrayLen == 0) {
-        return NULL;
-    }
+uint8_t opencoap_find_option(coap_option_iht* array, uint8_t arrayLen, coap_option_t option, uint8_t* startIndex) {
+   uint8_t i, j;
```
malishav: In OpenWSN, we typically declare one variable per line. Could you fix this throughout the code?

openapps/opencoap/opencoap.c
```diff
@@ -846,19 +873,23 @@ owerror_t opencoap_options_encode(
...
-    for (i = 0; i < arrayLen; i++) {
-        if (array[i].type == option) {
-            return &array[i];
-        }
-    }
-    return NULL;
+   for (i=0; i<MAX_COAP_OPTIONS; i++) {
+      if (array[i].type == option) {
+         // skip until different type
+         for (j=i; j<MAX_COAP_OPTIONS && array[j].type == option; j++) {}
```
malishav: I think it would be easier to read the code if you had a single loop and a counter incremented each time a condition is true. Would that not work?

openapps/opencoap/opencoap.c
```diff
@@ -276,48 +302,49 @@ void opencoap_receive(OpenQueueEntry_t* msg) {
...
-                   securityReturnCode = COAP_CODE_RESP_UNAUTHORIZED;
-               }
-               found = TRUE;
-            };
-         };
+      while (found==FALSE && securityReturnCode==COAP_CODE_EMPTY) {
+        
+        option_count = opencoap_find_option(coap_incomingOptions, coap_incomingOptionsLen, COAP_OPTION_NUM_URIPATH, &option_index);
+        if (
+             option_count >= 2                                       &&
```
malishav: Shouldn't this condition be true only if `option_count == 2`? I think there should be an `else` with the error handling if `option_count > 2`. Do you agree?

openapps/opencoap/opencoap.h
```diff
@@ -245,7 +245,7 @@ uint8_t opencoap_options_parse(uint8_t* buffer,
     uint8_t bufferLen,
     coap_option_iht* options,
     uint8_t* optionsLen);
-coap_option_iht* opencoap_find_option(coap_option_iht* array, uint8_t arrayLen, coap_option_t option); 
+uint8_t opencoap_find_option(coap_option_iht* array, uint8_t arrayLen, coap_option_t option, uint8_t* startIndex);
```
malishav: Why do you use `startIndex`? Would it not be possible to loop through the options and count the number of occurrences with only the first three parameters?

openapps/cinfo/cinfo.c
chris-ho: previously the cinfo reported the old stack version 1.9.0 after increasing that to 1.11.0 we need an additional character in the output string.

openapps/opencoap/opencoap.c
chris-ho: by today only path0/path1 is provided. i agree checking for `option_count == 2` in others cases no matching resource is found and we return a msg with `coap_header.Code = COAP_CODE_RESP_NOTFOUND;`

malishav:
@chris-ho let me know when you are done, so that I can do another pass and run some tests locally. The idea is to minimize the number of modified files by removing anything that is not specific to the issue.

chris-ho:
@malishav if you want me to seperate the changes from cinfo i can do so, otherwise the PR should be ready again.

malishav:
It seems your editor has changed the line endings in a bunch of files and you committed these changes to the repo. Before we can merge, we need to remove these edits. You can see them in the "Files changed" tab of the PR. Could you use git checkout to revert the changes to any file other than the ones in openapps. cjoin is fine.

chris-ho:
@malishav i'm a little bit confused because if i open some of the files you mentioned i have windows style line endings which is the format we are using right?

malishav:
We unfortunately have a mix of both unix line endings and windows. The problem for this PR is that you committed changes to files that should not be modified. You can see them [here](https://github.com/openwsn-berkeley/openwsn-fw/pull/369/files). An easy way to fix this is to checkout each individual file that should not be modified from the develop branch.

chris-ho:
cleanup handled in new branch https://github.com/openwsn-berkeley/openwsn-fw/pull/371

PR #370, opened 2017-07-20
FW-666 Easy channel hopping template modify 

andy199310:


twatteyne:
perfect, thanks a million!

PR #379, opened 2017-08-24
FW 693: printf command to send a string to openVisualizer

ftheoleyre:
by default, is active. Can be deactivated during the compilation by inserting printf=0 when compiling the firmware (e.g. scons  board=python toolchain=gcc printf=0 oos_openwsn)

If printf is deactivated, the corresponding code is not compiled for a matter of memory footprint (C macros #ifdef)

### DIFF
--- a/drivers/common/openserial.c
+++ b/drivers/common/openserial.c
@@ -23,6 +23,9 @@
 #include "icmpv6rpl.h"
 #include "icmpv6echo.h"
 #include "sf0.h"
+#include <stdio.h>
+
+//#define OPENSERIAL_PRINTF
 
 //=========================== variables =======================================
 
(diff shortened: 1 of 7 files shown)


drivers/common/openserial.c
```diff
@@ -245,6 +248,133 @@ owerror_t openserial_printSniffedPacket(uint8_t* buffer, uint8_t length, uint8_t
     return E_SUCCESS;
 }
 
+
+//===== printf and helper functions
+
+//append a uint8_t at the end of a string
+char *openserial_ncat_uint8_t(char *str, uint8_t val, uint8_t length){
```
changtengfei: what's this function used for?

drivers/common/openserial.c
```diff
@@ -245,6 +248,133 @@ owerror_t openserial_printSniffedPacket(uint8_t* buffer, uint8_t length, uint8_t
...
+   if (b != 0 || a != 0)
+       str[l++] = '0' + b;
+   str[l++] = '0' + c;
+   str[l++] = '\0';
+#endif
+   return(str);
+}
+
+//append a uint32_t at the end of a string (without the non significant zeros)
+char *openserial_ncat_uint32_t(char *str, uint32_t val, uint8_t length){
```
changtengfei: also here?

drivers/common/openserial.c
```diff
@@ -245,6 +248,133 @@ owerror_t openserial_printSniffedPacket(uint8_t* buffer, uint8_t length, uint8_t
...
+      val = val - power * digit;
+      power = power / 10;
+   }
+   str[l+shift] = '\0';
+#endif
+   return(str);
+}
+
+//append a uint32_t at the end of a string (without the non significant zeros)
+char *openserial_ncat_uint8_t_hex(char *str, uint8_t val, uint8_t length){
```
changtengfei: here also? I didn't see where this function is called

openstack/openstack.c
```diff
@@ -76,7 +76,12 @@ void openstack_init(void) {
    
    //===== applications
    openapps_init();
-   
+
```
changtengfei: This code is no necessary. When compiling, user already know printf is enabled. actually I think we can make it as default.

ftheoleyre:
obsolete pull request -> replaced with the strict minimum

PR #380, opened 2017-08-25
Fw 694: statistics through the serial line

ftheoleyre:
the firmware sends some stats to OV for a collection of events (DIO/DAO tx, packet reception and emission, 6P command (result, transmission, etc), schedule changes)


ftheoleyre:
more up to date pull request exists

PR #389, opened 2017-09-18
FW-704 - Assign to serial.baudrate instead of setBaudrate

sieben:
There are also some automatic trailing whitespaces fixes


sieben:
Perfect the CI pass 👍

PR #393, opened 2017-09-29
FW-681. Fixed handling of coap options

chris-ho:
this PR supports additional coap options to be used which have an index smaller than COAP_OPTION_NUM_URIPATH. It also handles multiple coap option of the same kind.

### DIFF
--- a/openapps/opencoap/opencoap.c
+++ b/openapps/opencoap/opencoap.c
@@ -108,6 +108,8 @@ void opencoap_receive(OpenQueueEntry_t* msg) {
    coap_option_iht           coap_outgoingOptions[MAX_COAP_OPTIONS];
    uint8_t                   coap_incomingOptionsLen;
    uint8_t                   coap_outgoingOptionsLen;
+   uint8_t                   option_count;
+   uint8_t                   option_index;
    owerror_t                 decStatus;
    coap_option_iht*          objectSecurity;
    coap_option_iht*          proxyScheme;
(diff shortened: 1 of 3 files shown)


openapps/opencoap/opencoap.c
```diff
@@ -276,48 +293,49 @@ void opencoap_receive(OpenQueueEntry_t* msg) {
...
+                memcmp(coap_incomingOptions[option_index+1].pValue,temp_desc->path1val,temp_desc->path1len)==0
+             ) {
+             if (temp_desc->securityContext != NULL && 
+                 blindContext != temp_desc->securityContext) {
+                 securityReturnCode = COAP_CODE_RESP_UNAUTHORIZED;
+             }
+              found = TRUE;
+          };
+
+        } else if (
```
malishav: Since `opencoap_find_option` is now generalized, it would be good if you could explicitly check `option_count == 1` here and also add a default `else` with error handling code in case an unsupported request arrives. Unsupported basically means 0 or >2 URIPATH options present.

openapps/opencoap/opencoap.c
```diff
@@ -849,19 +867,36 @@ owerror_t opencoap_options_encode(
...
-    if (array == NULL || arrayLen == 0) {
-        return NULL;
-    }
 
-    for (i = 0; i < arrayLen; i++) {
-        if (array[i].type == option) {
-            return &array[i];
+uint8_t opencoap_find_option(coap_option_iht* array, uint8_t arrayLen, coap_option_t option, uint8_t* startIndex) {
+   uint8_t i;
+   uint8_t j = 0;
```
malishav: A small detail, could you initialize `j` and `found` on a separate line. That's a convention that we generally use in OpenWSN.

malishav:
@chris-ho, thanks a bunch for fixing the line ending issue. I left two small comments, as soon as these are fixed, the PR should be ready to fly IMO.

chris-ho:
@malishav thanks for review

malishav:
thanks @chris-ho. tested the PR in simulation mode with cjoin app and with firefox copper plugin -- everything seems to be OK.  @changtengfei, my 👍 for this PR.

PR #400, opened 2017-11-27
FW-727. Fixes duplicate static variables in firmware images

jmrubillon:
Firmware images considered:
Telosb
openmotestm
iot-lab_M3


changtengfei:
This issue won't fix.

PR #401, opened 2017-11-27
FW-636. Port to DecaWave's ev1000 evaluation board

jmrubillon:
Working port to ev1000 evaluation board.
NOTE: Slot timing has been increased to 20ms as 110kbps on UWB is slower than on narrow-band due to preamble length and PAC symbols.


changtengfei:

PR #406, opened 2018-01-26
FW-693: printf capabilities for the openwsn motes

ftheoleyre:
firmware part (in openserial)


drivers/common/openserial.c
```diff
@@ -245,6 +248,65 @@ owerror_t openserial_printSniffedPacket(uint8_t* buffer, uint8_t length, uint8_t
...
+      return(str);
+
+   uint8_t  digit, shift, i;
+   uint32_t power;
+   bool     nonzero = FALSE;
+
+
+   power = 1000000000;
+   shift = 0;           // to avoid non significant zeros
+   for(i=0; i<10; i++){
```
changtengfei: Don't do the convert on mote side, leave it t othe openvisualizer do it.

PR #407, opened 2018-02-08
Fix "old-style function definition" warning all over the repo

PeterKietzmann:
This popped up during the OpenWSN integration into RIOT. There, warnings are treated as errors by default, which lead to tons of errors at first. What do you think about an update as proposed in this PR and how would you organize it? All in once commit, split up?

If it's not interesting to you, just close.


changtengfei:
@PeterKietzmann I think this is very cool! I will merge the pull request.

PR #408, opened 2018-02-08
Remove unneeded (?!) includes

PeterKietzmann:
As I see it, the application defendant includes are not needed, though I'm not entirely sure about it. @changtengfei can you please carefully check this?


changtengfei:
@PeterKietzmann those includes are required when the apps=applications options is used in compiling. The CI doesn't have compile with apps options by default.

PR #409, opened 2018-02-08
MAC: fix potentially uninitialized variable

PeterKietzmann:
This PR sets initial values to some variables. Compiler warnings complained about potential use of uninitialized variables.

### DIFF
--- a/openstack/02a-MAClow/IEEE802154.c
+++ b/openstack/02a-MAClow/IEEE802154.c
@@ -38,7 +38,7 @@ void ieee802154_prependHeader(OpenQueueEntry_t* msg,
    uint16_t     timeSyncInfo;
    uint16_t     length_elementid_type;
    bool         headerIEPresent = FALSE;
-   uint8_t      destAddrMode;
+   uint8_t      destAddrMode = -1;
    
    securityEnabled = msg->l2_securityLevel == IEEE154_ASH_SLF_TYPE_NOSEC ? 0 : 1;
 
--- a/openstack/02b-MAChigh/sixtop.c
+++ b/openstack/02b-MAChigh/sixtop.c
@@ -1061,9 +1061,9 @@ void sixtop_six2six_notifyReceive(
     uint8_t              length,
     OpenQueueEntry_t*    pkt
 ){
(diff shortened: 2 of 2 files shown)


changtengfei:
@PeterKietzmann the unassigned variables are variables to be determined in process. I don't think it's required to initialize them.

PeterKietzmann:
Yes, that's why they will be overwritten. However, there might be corner cases. At least the compiler complains about them being uninitialized.

changtengfei:
Peter, the compiler should be able detect, if it's defined inside of the function later, it shouldn't complaint.  Which compiler you are using, somehow I can compile the code without any warning.

PeterKietzmann:
I'm using arm-none-eabi-gcc 5.4.1. This problem occurred when  compiling OpenWSN as part of the integration in RIOT. There, we enable compiler warnings and treat them as errors. See this excerpt as an example, where I at least disabled to handle warnings as errors:

```
/media/Data/RIOT/tests/openwsn/bin/pkg/nucleo-f103/openwsn-fw/openstack/02a-MAClow/IEEE802154.c: In function 'ieee802154_prependHeader':
/media/Data/RIOT/tests/openwsn/bin/pkg/nucleo-f103/openwsn-fw/openstack/02a-MAClow/IEEE802154.c:175:7: warning: 'destAddrMode' may be used uninitialized in this function [-Wmaybe-uninitialized]
    if (destAddrMode == IEEE154_ADDR_SHORT) {
```

If you don't want this fix in your code, just close the. Then I need a proper patch as part of the RIOT integration.

changtengfei:
@PeterKietzmann I would like see the same what you saw with the warning. For now, is that possible for me to reproduce this warning?
Even finally we are going to have this fix, I would like to replace the -1 by enum type, which I can add them.

PeterKietzmann:
Fine. I'll push everything once #410 was merged. Will give some further details about reproducing it afterwards. We can close this PR then.

PeterKietzmann:
@changtengfei now that you have seen these warnings, what do you think about this PR?

changtengfei:
Thank you @PeterKietzmann !

PR #416, opened 2018-03-20
Fix 01bsp_cryptoengine test data

chris-ho:
when running CCMS_AUTH tests output is never compared because len_m is zero for both examples, use tag_len instead


malishav:
@changtengfei my +1 on this PR

changtengfei:
Thanks for fixing this!

PR #417, opened 2018-03-28
FW-740. Casting in opentimer.

ghost:
Operating with 16bit timer and compiling on 32bit architecture (ARM Cortex-M4) a default casting is done in variable instruction in opentimers.c so the overflow arithmetics does no work properly.

### DIFF
--- a/drivers/common/opentimers.c
+++ b/drivers/common/opentimers.c
@@ -356,9 +356,11 @@ void opentimers_timer_callback(void){
     PORT_TIMER_WIDTH timerGap;
     PORT_TIMER_WIDTH tempTimerGap;
     PORT_TIMER_WIDTH tempLastTimeout = opentimers_vars.currentTimeout;
+	opentimers_vars.running = FALSE;
     // 1. find the expired timer
     for (i=0;i<MAX_NUM_TIMERS;i++){
         if (opentimers_vars.timersBuf[i].isrunning==TRUE){
+			opentimers_vars.running = TRUE;
             // all timers in the past within TIMERTHRESHOLD ticks
             // (probably with low priority) will mared as Expired.
             if ((PORT_TIMER_WIDTH)(opentimers_vars.currentTimeout-opentimers_vars.timersBuf[i].currentCompareValue) <= TIMERTHRESHOLD){
@@ -369,9 +371,7 @@ void opentimers_timer_callback(void){
                     tempLastTimeout = opentimers_vars.timersBuf[i].currentCompareValue;
                 }
             }
-        } else if (i==(MAX_NUM_TIMERS-1)) {
-			opentimers_vars.running = FALSE;
-		}
+        }
     }
     
     // update lastTimeout
--- a/drivers/common/opentimers.c
+++ b/drivers/common/opentimers.c
@@ -7,6 +7,9 @@ at most MAX_NUM_TIMERS timers.
 \author Tengfei Chang <tengfei.chang@inria.fr>, April 2017.
  */
 
+#include <stddef.h>
+#include <string.h>
+
 #include "opendefs.h"
 #include "opentimers.h"
 #include "sctimer.h"
@@ -358,7 +361,7 @@ void opentimers_timer_callback(void){
         if (opentimers_vars.timersBuf[i].isrunning==TRUE){
             // all timers in the past within TIMERTHRESHOLD ticks
             // (probably with low priority) will mared as Expired.
-            if (opentimers_vars.currentTimeout-opentimers_vars.timersBuf[i].currentCompareValue <= TIMERTHRESHOLD){
+            if ((PORT_TIMER_WIDTH)(opentimers_vars.currentTimeout-opentimers_vars.timersBuf[i].currentCompareValue) <= TIMERTHRESHOLD){
                 // this timer expired, mark as expired
                 opentimers_vars.timersBuf[i].hasExpired = TRUE;
                 // find the fired timer who has the smallest currentTimeout as last Timeout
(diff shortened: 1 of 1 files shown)


drivers/common/opentimers.c
```diff
@@ -358,7 +361,7 @@ void opentimers_timer_callback(void){
         if (opentimers_vars.timersBuf[i].isrunning==TRUE){
             // all timers in the past within TIMERTHRESHOLD ticks
             // (probably with low priority) will mared as Expired.
-            if (opentimers_vars.currentTimeout-opentimers_vars.timersBuf[i].currentCompareValue <= TIMERTHRESHOLD){
+            if ((PORT_TIMER_WIDTH)(opentimers_vars.currentTimeout-opentimers_vars.timersBuf[i].currentCompareValue) <= TIMERTHRESHOLD){
```
changtengfei, line 365: I don't think this will make a difference with this changes. Both variable are PORT_TIMER_WIDTH type already.

drivers/common/opentimers.c
```diff
@@ -451,7 +454,7 @@ void opentimers_timer_callback(void){
                 // and its compare value is larger than timer "i" no more than TIMERTHRESHOLD ticks,
                 // replace candidate timer by timer "i".
                 if (opentimers_vars.timersBuf[i].priority < opentimers_vars.timersBuf[idToSchedule].priority){
-                    if (tempTimerGap - timerGap < TIMERTHRESHOLD){
+                    if ((PORT_TIMER_WIDTH)(tempTimerGap - timerGap) < TIMERTHRESHOLD){
```
changtengfei, line 458: it's the same here

mateusz-kubaszek:
Here is the problem reconstruction (the first number in log line is timestamp expressed in ms):
[log.txt](https://github.com/openwsn-berkeley/openwsn-fw/files/1859665/log.txt)
[opentimers.txt](https://github.com/openwsn-berkeley/openwsn-fw/files/1859670/opentimers.txt)
Please have a look at what is happening between 950ms and 1000ms.

changtengfei:
hi @mateusz-kubaszek I see. Could you indicating me the micro-controller you are using, which make you encounter this problem?

mateusz-kubaszek:
EZR32WG family. I use LETIMER0 as sctimer (it works on EM2 sleep mode, and the processor could be woken up in 2us with an interrupt).

drivers/common/opentimers.c
```diff
@@ -353,12 +356,14 @@ void opentimers_timer_callback(void){
     PORT_TIMER_WIDTH timerGap;
     PORT_TIMER_WIDTH tempTimerGap;
     PORT_TIMER_WIDTH tempLastTimeout = opentimers_vars.currentTimeout;
+	opentimers_vars.running = FALSE;
```
changtengfei: openwsn uses 4 spaces as indention to level the code. Could you fix it?

drivers/common/opentimers.c
```diff
@@ -353,12 +356,14 @@ void opentimers_timer_callback(void){
...
     // 1. find the expired timer
     for (i=0;i<MAX_NUM_TIMERS;i++){
         if (opentimers_vars.timersBuf[i].isrunning==TRUE){
+			opentimers_vars.running = TRUE;
```
changtengfei: indention problem.

changtengfei: also this check should be done at step 3 find the next timer, not here.

drivers/common/opentimers.c
```diff
@@ -461,9 +466,10 @@ void opentimers_timer_callback(void){
...
     
     // 4. reschedule the timer
-    opentimers_vars.currentTimeout = opentimers_vars.timersBuf[idToSchedule].currentCompareValue;
-    opentimers_vars.lastCompare[opentimers_vars.index] = opentimers_vars.currentTimeout;
-    opentimers_vars.index = (opentimers_vars.index+1)&0x0F;
-    sctimer_setCompare(opentimers_vars.currentTimeout);
-    opentimers_vars.running        = TRUE;
-}
\ No newline at end of file
+	if (TRUE==opentimers_vars.running){
```
changtengfei: please make sure the indention is correct and the following code.

changtengfei: Could you swipe the condition by putting the variable name on the left and the possible value on the right?

drivers/common/opentimers.c
```diff
@@ -7,6 +7,9 @@ at most MAX_NUM_TIMERS timers.
 \author Tengfei Chang <tengfei.chang@inria.fr>, April 2017.
  */
 
+#include <stddef.h>
```
changtengfei: I didn't see the requirement of this header

drivers/common/opentimers.c
```diff
@@ -7,6 +7,9 @@ at most MAX_NUM_TIMERS timers.
...
+#include <string.h>
```
changtengfei, line 10: also this one.

changtengfei (changes requested):
Hi @mateusz-kubaszek , I left some comments in your PR. Could you read them and do the fix?

drivers/common/opentimers.c
ghost: ok

drivers/common/opentimers.c
ghost: opentimers_init() uses memset which is declared in <string.h>

drivers/common/opentimers.c
ghost: Could you explain it in more detail?

changtengfei:

PR #419, opened 2018-04-11
Make leds_increment to handle overflow

luke0201:
While leds_circular_shift handles overflow, it seems that leds_increment doesn't, so I just copied the corresponding part of the source code.


changtengfei:

changtengfei:
No need to handle it.

PR #423, opened 2018-04-28
FW-760: Mimsy2 BSP 

skrillberg:
I've added a board support package for Mimsy2, the small inertial measurement mote we developed at


changtengfei:

PR #432, opened 2018-07-12
[FW-769] Fixes in OpenMote-B implementation

peretuset:
This PR is to fix firmware issues with the OpenMote-B board. In particular:
1) Make sure 2.4 GHz antenna switch works as expected
2) Make sure user button uses the proper pins
3) Remove unnecessary code (i.e., RevA1 is no longer needed since all boards use the 512-kbytes chip)


changtengfei:

changtengfei:
Most of the code are hold in FW-708 branch, part of the code has been merged from other PR already. 
As currently the multiple radio support is lead by Mina, we will not maintain this PR.

PR #433, opened 2018-07-24
FW-755. Add content format to join requests.

malishav:
This enables automated testign with F-Interop as Wireshark automatically decodes the application payload as CBOR. Otherwise, Content-Format is not needed.

### DIFF
--- a/openapps/cjoin/cjoin.c
+++ b/openapps/cjoin/cjoin.c
@@ -267,10 +267,16 @@ owerror_t cjoin_sendJoinRequest(open_addr_t* joinProxy) {
    // length and value are set by the CoAP library
 //   options[2].type = COAP_OPTION_NUM_OBJECTSECURITY;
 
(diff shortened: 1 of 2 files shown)


changtengfei:
@malishav I saw you added one option in the join request and I didn't see a PR on the openwsn-sw side. Does this option not requires to be processed on JRC side?

malishav:
Hmmm it doesn’t need to be processed as the JRC already knows that the
payload is CBOR. From what I remember, Wireshark will correctly dissect the
payload of the *response* as CBOR as well, even though the Content-Format
may not be present in the packet. But, let me double check this tomorrow
before merging..

malishav:
@changtengfei I opened a PR on the SW side:

https://github.com/openwsn-berkeley/openwsn-sw/pull/179

I tested with F-Interop develop branch and SECJOIN_01 and SECJOIN_02 tests pass.

PR #434, opened 2018-07-31
FW-767. Implementation of RS role of OSCORE profile of ACE framework.

malishav:
This PR adds support for a 3-party authorization by implementing OSCORE profile of ACE framework.

The role of RS is co-located with a 6TiSCH node, implemented as part of this PR.
The role of AS is co-located with JRC, implemented as part of openwsn-sw repository.
The role of Client is implemented as part of coap repository.


changtengfei:

PR #436, opened 2018-08-28
FW-737 

christianhopfner:
we found an issue which disabled the interrupts in the schedule and didn't reenabled them in all cases, that led to the stuck timers


changtengfei:
@chris-ho thanks for finding this bug! With this PR, am I able to close the FW-737? In another word, does it fix the issue described in FW-737?

christianhopfner:
@changtengfei it improves it but does not solve it completely

changtengfei:
Then there will be two fixes for this issue. I will merge the current one first. You can create another when you found the bug. Thanks!

PR #441, opened 2018-10-03
FW-778: Initial commit for the PKA drivers 

TimothyClaeys:
These first commits provide support for the ECDSA signature verification algorithm. Code is added to the cryptoengine files and the 01bsp_cryptoengine project has been updated with a test suite containing one signature verification over the secp256r1 curve.


changtengfei:

PR #457, opened 2019-02-15
Configure TX power through radio API

malishav:
This PR extends the `radio.h` API in order to provide support for configuring TX power. This is a necessary extension in order to enable an external agent, e.g. OpenVisualizer through CLI or OpenBenchmark platform, to configure the TX power of a given mote.

New `radio_setTxPower` API is implemented for CC2420, AT86RF231, and CC2538 -based boards. Mock up functions are added for other radio chips that are not supported at the moment.

The PR also removes the trailing whitespaces in all touched files.

### DIFF
--- a/bsp/boards/derfmega/radio.c
+++ b/bsp/boards/derfmega/radio.c
@@ -49,14 +49,14 @@ void radio_init(void) {
 
    // clear variables
    memset(&radio_vars,0,sizeof(radio_vars_t));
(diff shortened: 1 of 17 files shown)


malishav:
Another PR, part of an epic has been opened:

https://github.com/openwsn-berkeley/openwsn-fw/pull/470

making this PR obsolete.

PR #470, opened 2019-03-23
Firmware implementation of OpenBenchmark API v0.0.1.

malishav:
This PR implements the firmware components of the OpenBenchmark API and their interface with the OpenVisualizer. The PR:
- introduces a CoAP app called `cbenchmark` that can be triggered over serial to send packets and that reacts to the CoAP packets adhering to the OpenBenchmark API.
- Implements the RFC7967 CoAP No Response option in order to enable CoAP requests without a corresponding response being generated by the network
- Generation of asynchronous events over serial with the parameters as needed by OpenBenchmark platform
-Extends the radio.h API in order to provide support for configuring TX power. This is a necessary extension in order to enable an external agent, e.g. OpenVisualizer through CLI or OpenBenchmark platform, to configure the TX power of a given mote. New radio_setTxPower API is implemented for CC2420, AT86RF231, and CC2538 -based boards. Mock up functions are added for other radio chips that are not supported at the moment. 
- Adds support for CCM at iotlab-m3 nodes and their derivatives. It implements the CCM using hardware acceleration available in the AT86RF231 chip. While the CCM implementation passes all the unit tests available, its performance is bad due to the SPI transfer. For CCM to complete on a 127-byte frame, it takes ~5ms, making it impossible to use it for link-layer security processing. The provided CCM implementation enables the secure join protocol to complete, so it is therefore deemed useful.
- Removes the trailing whitespaces in all touched files. 

The PR resolves the following issues:
- https://openwsn.atlassian.net/browse/FW-806
- https://openwsn.atlassian.net/browse/FW-807
- https://openwsn.atlassian.net/browse/FW-829
- https://openwsn.atlassian.net/browse/FW-808
- https://openwsn.atlassian.net/browse/FW-703


changtengfei:

PR #472, opened 2019-03-31
FW-703. Implement cryptoengine for iotlab-m3 nodes.

malishav:
This PR adds support for CCM at iotlab-m3 nodes and their derivatives. It implements the CCM using hardware acceleration available in the AT86RF231 chip. While the CCM implementation passes all the unit tests available, its performance is bad due to the SPI transfer. For CCM to complete on a 127-byte frame, it takes ~5ms, making it impossible to use it for link-layer security processing.

The provided CCM implementation enables the secure join protocol to complete, so it is therefore deemed useful.


malishav:
Another PR, part of an epic, has been opened:

https://github.com/openwsn-berkeley/openwsn-fw/pull/470

making this PR obsolete.

PR #474, opened 2019-04-08
FW-775 clear sixtop state if ack is not received after max retries re…

christianhopfner:
…ached


changtengfei:
Thanks for fixing this issue!

PR #476, opened 2019-05-21
FW-841. Improve Mercator

yatch:
This PR implements the items listed in FW-841.

Once these changes are merged, we need to update wiki pages on Mercator:
* https://github.com/openwsn-berkeley/mercator/wiki/Mercator-Firmware
* https://github.com/openwsn-berkeley/mercator/wiki/How-to-build-the-firmware

The code is tested with FIT/IoT-LAB (`iot-lab_M3`) M3 motes, and with OpenMote-B motes (`openmote-b-24ghz`).

### DIFF
--- a/projects/common/03oos_mercator/03oos_mercator.c
+++ b/projects/common/03oos_mercator/03oos_mercator.c
@@ -127,6 +127,9 @@ void serial_tx_RESP_ST(void) {
 }
 
 void serial_rx_REQ_IDLE(void) {
+   // schedule a response
+   scheduler_push_task(serial_tx_RESP_IDLE,TASK_PRIO_SERIAL);
+
    if (mercator_vars.uartbufrxindex!=sizeof(REQ_IDLE_ht)){
       // update stats
       mercator_vars.serialNumRxWrongLength++;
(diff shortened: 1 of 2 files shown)


bsp/boards/iot-lab_M3/uart.c
```diff
@@ -95,12 +95,9 @@ void uart_clearTxInterrupts(void) {
 }
 
 void uart_setCTS(bool state) {
-    
-    if (state==0x01) {
-        USART_SendData(USART1,(uint16_t)XON);
-    } else {
-        USART_SendData(USART1,(uint16_t)XOFF);
-    }
+    // do nothing since the serial connection is bridged by a TCP
```
changtengfei: With this changes, the M3 won't be able to work locally. Is there a problem if using the current implementation on FIT/IoT-LAB?

yatch: XON/XOFF will be transmitted to user's terminal if I understand it correctly. If you want to use M3 locally, disabling the software flow control is one of solutions... :-/

Do we have a compilation option to disable XON/XOFF? ;-)

yatch: Thank you for the review; I think I had a wrong idea. I confirmed that the mercator firmware works without this change. I will revert this.

changtengfei: Thanks so much for taking the comments!

PR #480, opened 2019-06-11
FW-703. Cryptoengine implementation for iot-lab M3 nodes.

malishav:
This PR adds support for CCM at iotlab-m3 nodes and their derivatives. It implements the CCM using hardware acceleration available in the AT86RF231 chip. While the CCM implementation passes all the unit tests available, its performance is bad due to the SPI transfer. For CCM to complete on a 127-byte frame, it takes ~5ms, making it impossible to use it for link-layer security processing.
The provided CCM implementation enables the secure join protocol to complete, so it is therefore deemed useful.

### DIFF
--- a/bsp/boards/iot-lab_A8-M3/SConscript
+++ b/bsp/boards/iot-lab_A8-M3/SConscript
@@ -23,6 +23,12 @@ source   = \
     Glob('configure/*.s') + \
     Glob('library/src/*.c')
 
+localEnv.Append(
+        CPPPATH =  [
+            os.path.join('#','bsp','chips','at86rf231'),
+        ],
+)
+
 board  = localEnv.Object(source=source) + rf231
 
 Return('board')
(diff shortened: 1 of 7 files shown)


changtengfei:
@malishav For the files:
- iotlab-m3.cfg
- iotlab-m3-bsl.py
Does the original configuration not work for you?  I know, I think part of the configuration is supporting the GDP debugging, now it's removed.  Do you intend to do so?

malishav:
When I worked on this implementation two years ago, I wasn't able to install the openocd with the ft2232 interface as it was reported obsolete. I replaced the configuration with RIOT's config and ftdi interface and was able to flash the motes with it. I haven't tested if this config works with GDB.

malishav:
@changtengfei I removed the commit modifying the bootloader configuration for Iot-lab M3 ndoes as I wans't able to get it working again on Windows. Let's keep using the old config for now and if the need arises we can revert to the new cofnig using the ftdi interface.

changtengfei:
Great! Thanks for the testing!

PR #486, opened 2019-09-17
FW-856: security_isConfigured() for DAGroot to enable EB 

ftheoleyre:
We assume the dagroot has the keys and is ready after the initialization.

Else, IEEE802154_security_isConfigured(void) always returns FALSE, and the Dagroot is never able to send an EB. As a consequence, no other device can join the network (no possibility to be synced).

I'm not a security expert, and I assume that the keys are correctly configured after the initialization of the DAGROOT (in the openapp cjoin.c)


ftheoleyre:
OpenVisualizer is now a requirement to set the security keys when enabling a DAG root.

PR #496, opened 2020-01-31
scons: add disable_ch_hop option

fjmolinas:
This PR adds a debugging option to disable channel hopping. I've found myself doing this often to speed up synchronization, joining, etc. while developing or debugging. The option disabling channel hopping takes the channel to be set as a parameter:

e.g.:

`scons single_ch_slot=26 board=openmote-b-24ghz bootload=/dev/riot/tty-openmote-b_1 toolchain=armgcc oos_openwsn
`

### DIFF
--- a/SConscript
+++ b/SConscript
@@ -48,6 +48,8 @@ if env['noadaptivesync']==1:
     env.Append(CPPDEFINES    = 'NOADAPTIVESYNC')
 if env['l2_security']==1:
     env.Append(CPPDEFINES    = 'L2_SECURITY_ACTIVE')
+if env['fix_channel']>=11 and env['fix_channel']<=26:
+    env.Append(CPPDEFINES    = {'IEEE802154E_SINGLE_CHANNEL' : env['fix_channel']})
 if env['msf_adapting_to_traffic']==1:
     env.Append(CPPDEFINES    = 'MSF_ADAPTING_TO_TRAFFIC')
 if env['printf']==1:
--- a/SConstruct
+++ b/SConstruct
@@ -84,6 +84,8 @@ project:
     printf        Sends the string messages to openvisualizer  
                   0 (off ), 1 (on, default)
     ide           qtcreator
+    fix_channel   Set single channel hopping for debugging
+                  0 (off, default), i (on, channel=i [11:26])
 
     Common variables:
     verbose        Print each complete compile/link command.
@@ -147,6 +149,7 @@ command_line_options = {
     'atmel_24ghz':              ['0','1'],
     'noadaptivesync':           ['0','1'],
     'l2_security':              ['0','1'],
+    'fix_channel':              ['0'] + map(str, range(11, 27)),
     'msf_adapting_to_traffic':  ['0','1'],
     'printf':                   ['1','0'],          # 1=on (default),  0=off
     'deadline_option':          ['0','1'],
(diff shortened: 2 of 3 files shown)


TimothyClaeys:
I think this is a nice command line option to have. Personally, the first thing I do when I start debugging is disabling the channel hopping. The only thing I would change is the name of the option. I think `single_ch_slot` sounds confusing, maybe change it to `disable_ch_hop` or `fix_channel`. What do you think @changtengfei.

SConscript
```diff
@@ -48,6 +48,8 @@ if env['noadaptivesync']==1:
     env.Append(CPPDEFINES    = 'NOADAPTIVESYNC')
 if env['l2_security']==1:
     env.Append(CPPDEFINES    = 'L2_SECURITY_ACTIVE')
+if env['single_ch_slot']!=0:
```
changtengfei: Need to guarantee the value is valid, such as 11-26

SConstruct
```diff
@@ -147,7 +149,8 @@ command_line_options = {
     'atmel_24ghz':              ['0','1'],
     'noadaptivesync':           ['0','1'],
     'l2_security':              ['0','1'],
-    'msf_adapting_to_traffic':  ['0','1'],
+    'single_ch_slot':           ['0'],
```
changtengfei: give the options as indicated in the help comments: 
such as, 0, 11-26

changtengfei (commented):
Hi @fjmolinas I like this option. I left some comments which are easy to fix. 
I agree with @TimothyClaeys to change to another name for the option.

SConscript
fjmolinas: That would only apply for 2.4 GHz, is that OK?

TimothyClaeys: Yeah, I think that is fine, we're currently only operating in the 2.4 Ghz band. I don't think the subGhz radio already works on the OpenMote-B's.

fjmolinas:
> single_ch_slot sounds confusing, maybe change it to disable_ch_hop or fix_channel. What do

@TimothyClaeys @changtengfei Is only the scons option confusing or also the define? I agree `disable_ch_hop` being a better command line option. If the define is also confusing maybe `IEEE802154E_SINGLE_CHANNEL`?

TimothyClaeys:
Yes, I would also rename the define.

fjmolinas:
@TimothyClaeys @changtengfei I addressed your comments, let me know if its ok to squash!

SConscript
```diff
@@ -48,6 +48,8 @@ if env['noadaptivesync']==1:
     env.Append(CPPDEFINES    = 'NOADAPTIVESYNC')
 if env['l2_security']==1:
     env.Append(CPPDEFINES    = 'L2_SECURITY_ACTIVE')
+if env['disable_ch_hop']!=11 and env['disable_ch_hop']<=26:
```
changtengfei: Could be better using `fix_channel` instead of `disable_ch_hop`? It's strange to have `disable_ch_hop` to have a value.

Also the checking condition is wrong.

changtengfei (commented):
Hi @fjmolinas Just one comment left.  Could you resolve it and then squash? After you resolve it I will merge it.

SConscript
fjmolinas: Ups, Sorry that was typo that got in when re-writing. addessed.

fjmolinas:
> Hi @fjmolinas Just one comment left. Could you resolve it and then squash? After you resolve it I will merge it.

Addresed!

TimothyClaeys (approved):

changtengfei (approved):

changtengfei:
@fjmolinas just merged!  Very nice to have this feature!

PR #506, opened 2020-04-21
bsp/boards/stm32f1/sctimer: fix time division

fjmolinas:
This PR fixes a bug in the time division implementation for stm32f1 boards,  `sctimer_vars.convert` is never set to false, this error would only show up every 2^31/16384.

### DIFF
--- a/bsp/boards/iot-lab_A8-M3/sctimer.c
+++ b/bsp/boards/iot-lab_A8-M3/sctimer.c
@@ -41,6 +41,7 @@ sctimer_vars_t sctimer_vars;
 void sctimer_init(void) {
     // clear local variables
     memset(&sctimer_vars,0,sizeof(sctimer_vars_t));
+    sctimer_vars.convert = TRUE;
     //enable BKP and PWR, Clock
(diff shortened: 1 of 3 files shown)


fjmolinas:
@changtengfei could you take a look? Should be straight forward since we discussed this offline.

changtengfei:
@fjmolinas thanks for fixing this issue. Have you tested on one of the boards that is able to form a network? Just trying to make sure it doesn't break anything.

fjmolinas:
> @fjmolinas thanks for fixing this issue. Have you tested on one of the boards that is able to form a network? Just trying to make sure it doesn't break anything.

I tested for a short time, I'll lunch an experiment running for a long time and report back.

fjmolinas:
> I tested for a short time, I'll lunch an experiment running for a long time and report back.

I also wasn't testing on a pure OpenWSN network, so I'll do that as well.

fjmolinas:
Actually seems like although the change makes sense code wise, it breaks something since I lost proper operation of the network... I need to revisit.

PR #520, opened 2020-08-21
openstack/02b-MAChigh/neighbors: fix neighbors_isNeighborWithHigherDagGrank

fjmolinas:
According to the api:

`returns TRUE if that neighbor is in use and has a higher DAG rank than me, FALSE otherwise.`

Function is currently returning True if they have the same DAG rank.

### DIFF
--- a/openstack/02b-MAChigh/neighbors.c
+++ b/openstack/02b-MAChigh/neighbors.c
@@ -246,7 +246,7 @@ bool neighbors_isNeighborWithHigherDAGrank(uint8_t index) {
     bool returnVal;
 
     if (neighbors_vars.neighbors[index].used == TRUE &&
-        neighbors_vars.neighbors[index].DAGrank >= icmpv6rpl_getMyDAGrank()) {
+        neighbors_vars.neighbors[index].DAGrank > icmpv6rpl_getMyDAGrank()) {
         returnVal = TRUE;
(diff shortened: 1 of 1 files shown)


fjmolinas:
@TimothyClaeys @changtengfei can maybe review this one?

fjmolinas:
> Function is currently returning True if they have the same DAG rank.

Maybe this is what is expect form the function and only the api needs clarification

fjmolinas:
BTW tested and everything seems to work fine, only nodes with same rank are not advertised as children.

```
11:54:18 INFO received RPL DAO from bbbb:0:0:0:56ad:fc65:106b:1114
        - parents:
           bbbb:0:0:0:26bc:fb65:106b:1114
        - children:
11:54:57 INFO received RPL DAO from bbbb:0:0:0:36bc:fd65:106b:1114
        - parents:
           bbbb:0:0:0:26bc:fb65:106b:1114
        - children:
11:55:09 INFO received RPL DAO from bbbb:0:0:0:36bc:fd65:106b:1114
        - parents:
           bbbb:0:0:0:26bc:fb65:106b:1114
        - children:
```

fjmolinas:
This one should be a low hanging fruit :)

fjmolinas:
Thanks for the review!

PR #521, opened 2020-09-01
kernel: move scheduler vars definitons to scheduler_types.h

fjmolinas:
This PR adds a `scheduler_types.h` header to allow custom definitions for `scheduler_vars_t` and `scheduler_dbg_t`. This way it is easy to implement the scheduler with an `RTOS` by simply implementing `scheduler.h` and having specific definitions for  `scheduler_vars_t` and `scheduler_dbg_t` based on the implementation.

See: https://github.com/RIOT-OS/RIOT/pull/14905 for an examples

### DIFF
--- a/SConscript
+++ b/SConscript
@@ -1141,6 +1141,7 @@ def buildIncludePath(project_dir, local_env):
             CPPPATH=[
                 os.path.join('#', 'inc'),
                 os.path.join('#', 'kernel'),
+                os.path.join('#', 'kernel', 'openos'),
                 os.path.join('#', 'openstack'),
                 os.path.join('#', 'openstack', '02a-MAClow'),
                 os.path.join('#', 'openstack', '02b-MAChigh'),
--- a/drivers/SConscript
+++ b/drivers/SConscript
@@ -49,6 +49,7 @@ else:
     localEnv.Append(
         CPPPATH =  [
             os.path.join('#','drivers','common'),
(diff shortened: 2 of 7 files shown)


TimothyClaeys:
@fjmolinas Could you change the header guard to OPENWSN_SCHEDULER_TYPES_H

fjmolinas:
@TimothyClaeys pushed and squashed, also I had missed adding the new header to the sconscript, lets see if travis is happy now.

fjmolinas:
hmm not enough

fjmolinas:
I don't think I quite understand how scons work... can't get it to see the new header.. :(

fjmolinas:
THanks for the help @TimothyClaeys I squashed both commits.

TimothyClaeys:
@fjmolinas One last remark before I merge. Is there a reason why this include is at the bottom of the `scheduler.h` file?

`#include "scheduler_types.h"`

fjmolinas:
> @fjmolinas One last remark before I merge. Is there a reason why this include is at the bottom of the `scheduler.h` file?
> 
> `#include "scheduler_types.h"`

Yes, it was to avoid circular dependencies between the headers. Which I now realize was still the case, but is addressed with the latest commit (push).

fjmolinas:
Thanks a lot for the review and the help @TimothyClaeys!

PR #523, opened 2020-09-02
kernel/scheduler: cleanups, optional debug vars

fjmolinas:
This PR adds a configuration to enable scheduler debug variables `SCHEDULER_DEBUG_ENABLE` (current and peak tasks). Disabling it will save a couple of bytes in memory.

This PR also adds geters for those variables to avoid reading them through `scheduler_dbg_t`

`numTasksCur` and `numTasksMax` are removed from `scheduler_vars_t` since there was no use of them.


TimothyClaeys:
@fjmolinas, looks good, thanks!

fjmolinas:
Thanks for the reivew!

PR #530, opened 2020-10-26
openstack/02b-MAChigh/msf.h: add ifndef guards

fjmolinas:
This PR simply adds some ifndef guards to allow easily modifying the values of the msf algorithm.

### DIFF
--- a/inc/config.h
+++ b/inc/config.h
@@ -312,11 +312,26 @@
  *
  * Allow the MSF algorithm to dynamically remove and allocate slots, based on the traffic load in the network.
  *
+ * Configuration options:
+ *  - MSF_MAX_NUMCELLS: cell usage is calculated every MSF_MAX_NUMCELLS elapsed cells.
+ *  - MSF_LIM_NUMCELLSUSED_HIGH: if cell usage is above this value, trigger 6P to add a single cell to the selected parent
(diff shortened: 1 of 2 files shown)


fjmolinas:
Maybe @TimothyClaeys can take a look?

TimothyClaeys:
@fjmolinas I think it is a good idea to make it easy to customize these values during compile time. Maybe you can add the ifdef guards to the `config.h` file, similar to how the values for `IEEE802154E_SINGLE_CHANNEL` or `PACKETQUEUE_LENGTH` are set.

TimothyClaeys (changes requested):
Maybe add:

```
#ifndef MSF_MAX_NUMCELLS
#define MSF_MAX_NUMCELLS                   32
#endif
```
to the `config`.h file and replace the defines here with:
```
#ifndef MSF_MAX_NUMCELLS
#define MAX_NUMCELLS                   32
#else
#define MAX_NUMCELLS                   MSF_MAX_NUMCELLS 
#endif
```
In the `config.h` these values only are useful when `ADAPTIVE_MSF` is active. So maybe add an `#if` guard like we did for the `OPENWSN_6LO_FRAGMENTATION_C` options

fjmolinas:
Done, lets see what travis says

fjmolinas:
@TimothyClaeys all green here!

TimothyClaeys (approved):

fjmolinas:
Thanks of the review!

PR #531, opened 2020-10-27
openapps/cjoin: add missing opentimers include

fjmolinas:
Opentimer types are used in the defintion of `cjoin_vars_t` but the include is missing, this PR fixes this.


fjmolinas:
Maybe @TimothyClaeys can take a look at this one

fjmolinas:
Thanks for the review!

PR #532, opened 2020-10-28
inc/config: use channel hopping by default

fjmolinas:
Not sure what maintainers prefer here, but maybe by default channel hopping should be enabled. I think it is what someone would expect when running the default build for OpenWSN. Feel free to close this one if you disagree.


fjmolinas:
ping @TimothyClaeys !

fjmolinas:
Thanks for the review!

PR #539, opened 2020-12-04
inc/opendefs: add new neighbors logging

fjmolinas:
This PR adds some information about the new neighbor RSSI, it was useful for me to realize that
nodes at the same distance where hearing the same nodes with a huge RSSI penalty.

### DIFF
--- a/inc/opendefs.h
+++ b/inc/opendefs.h
@@ -234,11 +234,12 @@ enum {
    // l2b
    ERR_SCHEDULE_ADD_DUPLICATE_SLOT     = 0x20, // the slot {0} to be added is already in schedule
    ERR_NEIGHBORS_FULL                  = 0x21, // neighbors table is full (max number of neighbor is {0})
+
    ERR_NO_SENT_PACKET                  = 0x22, // there is no sent packet in queue
    ERR_NO_RECEIVED_PACKET              = 0x23, // there is no received packet in queue
    ERR_SCHEDULE_OVERFLOWN              = 0x24, // schedule overflown
    ERR_SIXTOP_RETURNCODE               = 0x25, // sixtop return code {0} at sixtop state {1}
-   ERR_SIXTOP_REQUEST                  = 0x26, // sending a 6top request
+   ERR_SIXTOP_REQUEST                  = 0x26, // sending a 6top request {0}
    ERR_SIXTOP_COUNT                    = 0x27, // there are {0} cells to request mote
    ERR_SIXTOP_LIST                     = 0x28, // the cells reserved to request mote contains slot {0} and slot {1}
    ERR_UNSUPPORTED_FORMAT              = 0x29, // the received packet format is not supported (code location {0})
--- a/inc/opendefs.h
+++ b/inc/opendefs.h
@@ -234,7 +234,6 @@ enum {
    // l2b
    ERR_SCHEDULE_ADD_DUPLICATE_SLOT     = 0x20, // the slot {0} to be added is already in schedule
    ERR_NEIGHBORS_FULL                  = 0x21, // neighbors table is full (max number of neighbor is {0})
-
    ERR_NO_SENT_PACKET                  = 0x22, // there is no sent packet in queue
    ERR_NO_RECEIVED_PACKET              = 0x23, // there is no received packet in queue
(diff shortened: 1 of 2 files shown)


fjmolinas:
ping @TimothyClaeys :)

fjmolinas:
Do you think this might be of interest? The default logging I can change and guard the rest depending on the logging level.

fjmolinas:
Maybe @malishav could take a look?

inc/opendefs.h
```diff
@@ -238,7 +238,7 @@ enum {
    ERR_NO_RECEIVED_PACKET              = 0x23, // there is no received packet in queue
    ERR_SCHEDULE_OVERFLOWN              = 0x24, // schedule overflown
    ERR_SIXTOP_RETURNCODE               = 0x25, // sixtop return code {0} at sixtop state {1}
-   ERR_SIXTOP_REQUEST                  = 0x26, // sending a 6top request
+   ERR_SIXTOP_REQUEST                  = 0x26, // sending a 6top request {0}
```
malishav, line 241: is this an unrelated bug fix?

SConstruct
```diff
@@ -124,7 +124,7 @@ command_line_options = {
         'armgcc',
         'gcc',
     ],
-    'logging': [str(l) for l in range(6)],
+    'logging': [str(l) for l in range(7)],
```
malishav: I don't understand why do you increase the potential logging level value here?

SConstruct
```diff
@@ -186,7 +186,7 @@ command_line_vars.AddVariables(
     (
         'logging',  # key
         '',  # help
-        command_line_options['logging'][5],  # default
+        command_line_options['logging'][6],  # default
```
malishav: see comment above

openstack/02b-MAChigh/neighbors.c
```diff
@@ -681,6 +682,14 @@ void registerNewNeighbor(open_addr_t *address,
                 if (rssi < GOODNEIGHBORMINRSSI) {
                     break;
                 }
+                open_addr_t addr_16;
+                addr_16.type = ADDR_16B;
+                packetfunctions_mac64bToMac16b(address, &addr_16);
+                uint16_t short_addr = ((addr_16.addr_16b[0] << 8) & 0xff00) |
```
malishav: I was thinking here whether we did this sort of conversion from `ADDR_16B` type to a `uint16` anywhere else in the code but did not find a similar use case. a short address is logged for `ERR_NO_NEXTHOP` but that implementation requires two arguments which is not applicable to your case since you want to log both the short address and the RSSI. With that in mind, I am ok with this approach.

openstack/02b-MAChigh/neighbors.c
```diff
@@ -681,6 +682,14 @@ void registerNewNeighbor(open_addr_t *address,
                 if (rssi < GOODNEIGHBORMINRSSI) {
                     break;
                 }
+                open_addr_t addr_16;
```
malishav: the convention we use is to declare variables at the beginning of the function, one var per line. applies throughout.

malishav (changes requested):
A couple of comments inline.

inc/opendefs.h
fjmolinas: Its in its own commit but, yes its unrelated, do you want me to split it out?

openstack/02b-MAChigh/neighbors.c
fjmolinas: So you want me to move it up to line 669? AFAIU putting it inside the iff reduces its scope to where it's needed, but I can move it up if you prefer.

SConstruct
fjmolinas: I'll remove

SConstruct
fjmolinas: I changed the default loggin to verbose, but I can remove that commit

fjmolinas: removed

openstack/02b-MAChigh/neighbors.c
fjmolinas: Moved up as suggested

malishav (approved):

malishav:
@fjmolinas thanks for doing the changes. Before I merge, could you rebase this on top of the latest develop and confirm what you have tested locally? The failing build seems unrelated.

PR #540, opened 2020-12-04
openapps/uinject/uinject: fix ticks calculation

fjmolinas:
This PR fixes a bug introduced during the sock migration, where I did not pass a pointer reference.


fjmolinas:
ping @TimothyClaeys this one should be a low hanging fruit!

PR #543, opened 2020-12-04
inc/config: make coap port configurable

fjmolinas:
This PR wants to make the default COAP configurable. Needed uit when using both RIOT's coap and OpenWSN since they both default to 5683 (RIOT's default is already configurable but would be nice to have this one as well.)


fjmolinas:
ping @TimothyClaeys :)

TimothyClaeys (approved):

PR #544, opened 2020-12-04
openstack/04-TRAN: move sock to own directory

fjmolinas:
In RIOT we finally went with hosting the patched version of OpenWSN in our sourcetree. To avoid  header and source file conflicts the easier for me was to move all sock files to its own directory, If this approach is not an issue for you it would make our life easier :)


fjmolinas:
ping @TimothyClaeys :)

